PR #41, opened 2023-03-15
Introduce support for Select entities, and add use_mode

canton7:
First off, I got a bit carried away on my day off. I'll be quite understanding if you don't want to accept this!

I wanted to be able to see and change the Work Mode (since it's an agent-only setting, which is settable through modbus). One thing led to another, and...

image

I've tried to split it into fairly independent commits -- it should be easiest to review them one-by-one.


nathanmarlor (approved):

nathanmarlor:
@canton7 - absolutely fantastic change, thank you!

canton7:
Thank you!

I broke your lints though... I'll get in a PR to fix those in a min

canton7:
Thank you!

I broke your lints though... I'll get in a PR to fix those in a min

PR #46, opened 2023-03-15
Add ModbusNumber, and use to control min/max SoC

canton7:
Follow the same approach as for the Select.

I couldn't figure out how to convince HA to show a battery as the icon:

image

Do you have any ideas?


nathanmarlor (approved):

nathanmarlor:
Thanks! Merged and added icons separately :)

canton7:
Aha, nice!

I was trying to icon sets working, so it will show a battery icon with the right level in it... But your way is probably cleared, and definitely simpler!

nathanmarlor:
Not a bad idea but it'd always show the min SoC icons in red with an exclamation mark 😂 just thought - probably want to keep the old sensors in there in case people have automations etc. setup (the domain will have changed, sensor -> number). Just nipped out but will add them back in later.

PR #52, opened 2023-03-17
Be smarter about which sensors to notify after a poll

canton7:
Most of the time, most sensors aren't changing. Telling HA that everything has changed after every poll is a bit wasteful, particularly if the user has the poll rate set high.

Make sure that we only tell HA about registers which have actually changed.


nathanmarlor:
:rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket:

PR #53, opened 2023-03-17
Refactor to introduce centralised knowledge of inverter registers and capabilities

canton7:
This is quite a big change, sorry! I didn't quite know where it was going to end up when I started, and I quite understand if you don't want to take it. Also very happy to take a review and make changes.

The motivation is twofold:

1. Generating an optimal sequence of register reads for a particular interver. This means knowing all of the address of all registers on a particular inverter
2. Writing force charge windows is probably going to need some centralised knowledge about an inverter (such as which registers need to be written in one block), and currently the knowledge about an inverter is spread across the codebase

The central points are inverter_connection_types, a list of ways to connect to inverters, and and inverter_profiles, a list of all inverter models which are supported, along with inverter-model-specific metadata, and the list of entities which are supported by that inverter model.

All parts of the codebase which need to deal with the differences between inverter models and connection types now go through these central points.

The `address_ranges` members will go when I get the address read grouping in place. I'll only bother with that if you want to accept this however (or something else, which means that we have single list of all addresses for a particular inverter)

### DIFF
--- a/custom_components/foxess_modbus/__init__.py
+++ b/custom_components/foxess_modbus/__init__.py
@@ -8,8 +8,6 @@ import asyncio
 import logging
 
 import voluptuous as vol
-from custom_components.foxess_modbus.const import FRIENDLY_NAME
-from custom_components.foxess_modbus.modbus_client import ModbusClient
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import Config
 from homeassistant.core import HomeAssistant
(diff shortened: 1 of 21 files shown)


nathanmarlor:
You beat me to it :rocket: 

Knew I'd made things a little worse when I added the latest models, but was short on time... Ideally wanted to hold all inverter 
config in a centralised model and inject it into the controller, which is exactly what you've done :+1: 

Thank you, you're on :fire: :raised_hands:

nathanmarlor:
@canton7 just a thought - up to you if you want to give it a shot, happy to pick it up over the weekend - but it'd be cool if we could introduce inverter (maybe connection type) specific poll/max reads into the centralised config? Not sure how that'd play with the options config flow, needs some thought.

canton7:
Yeah I was wondering about that. At the least, I think it would make sense to set a defaults for an inverter type, even if we then let the user tweak them still.

I'll see whether I get around to it today -- currently going down a rabbit hole trying to speed up this W610 some more!

PR #55, opened 2023-03-17
Be smarter about grouping address into read commands

canton7:
For a H1 AUX with a read size of 8, this reduces the number of reads from 9 to 7, although it's not optimal in the general case.
    
It also avoids having to hard-code the address range for each inverter model.


nathanmarlor:
:fire:

PR #58, opened 2023-03-18
Part 1 of properly supporting charge windows

canton7:
In the interface, this:

1. Adds a new sensor "Period X - Enable force charge", which mirrors the foxess app regarding whether force charge is enabled
2. Tweaks the "Period X - Start" and End sensors so that they remember their last values when force-charging is disabled

Together, these mean we have 4 sensors for each charge period which behave like the app. If the user creates a card to control time periods (or we create one for them), we'll remember the window they configured even when enable force charge is disabled.

Internally, pulls together the 3 sensors relating to time windows into a common object, which will be needed when we come to write a service to set them together.


custom_components/foxess_modbus/entities/modbus_time_period_sensors.py
```diff
@@ -0,0 +1,138 @@
...
+        """Return specific state data to be restored."""
+        return RestoredExtraData(
+            json_dict={"last_non_zero_value": self._last_non_zero_value}
+        )
+
+    def update_callback(self, changed_addresses: set[int]) -> None:
+        """Schedule a state update."""
+        if self.entity_description.address in changed_addresses:
+            value = self._controller.read(self.entity_description.address)
+            if value is not None and value > 0:
```
canton7, line 83: This isn't quite right, as it's valid to set a start time of midnight. We need to check the other sensor as well.

canton7:
Hold off on this for now - spotted an issue

canton7:
All fixed!

canton7:
I'm in two minds about renaming `time_period_x_enabled`. On the one hand that doesn't really indicate that it's "enable change from grid". On the other hand, I don't like breaking changes..

canton7:
Hold off on this again -- needs an update to take account of #55

nathanmarlor:
> I'm in two minds about renaming `time_period_x_enabled`. On the one hand that doesn't really indicate that it's "enable change from grid". On the other hand, I don't like breaking changes... Any thoughts?

Just tested the changes - looks great. I think eventually we'll need to move away from the original description set so I'm not too concerned about breaking changes - it's not a huge job to update/rename sensors and the updates aren't forced. Thoughts?

nathanmarlor:
> Thanks! I think renaming is fine -- I'm wondering about changing the entity_id (sorry, wasn't particularly clear there)

My bad - I actually meant both description and entity_id. FYI can you update the descriptions to be 'proper case' please?

canton7:
(It might take me a day of two to find time for this, just to set expectations)

nathanmarlor:
No worries, no rush 🙂

nathanmarlor:
:100: :100: :100: :100: :100: :100: :100: :100: 

Thanks :+1:

PR #79, opened 2023-03-24
Added battery charge/discharge sensors

nathanmarlor:


custom_components/foxess_modbus/entities/xx1_aux_sensors.py
```diff
@@ -82,6 +82,42 @@
         native_unit_of_measurement="kWh",
         scale=0.1,
     ),
+    SensorDescription(
+        key="charge_total",
```
canton7, line 86: @nathanmarlor The sensors for current power are called `battery_charge` and `battery_discharge`. Would it be worth calling these `battery_charge_total` etc to match?

PR #84, opened 2023-03-25
Remove the need for separate lists of entity types

canton7:
Remove the need for separate lists of entity types

I've been wanting to do this for a while. Instead of having to keep separate lists of the different types of entities, which is only going to grow when we start adding integrals and other things, teach each EntityDescription how to create an instance of its corresponding Entity.

This means we can mash everything into one big list of entity descriptions, and when HA asks for all binary sensors we find the ones which will create binarys sensors, and ask them to create those.

Still at draft, as I still want to sort the list of entites by register address. But I wanted to get your feedback and see whether this was a direction you wanted to go in.


nathanmarlor:
Looks great...! Certainly less boilerplate needed when adding additional sensor types in the future. :smiley:

canton7:
I forgot I left this on draft... I'll take your comment as approval and merge it myself!

Issue #85, opened 2023-03-26
Daylight savings time-related weirdness?


canton7:
## Version of the custom_component

v1.4.0b3

## Describe the bug

So, this morning I had a fairly large spike of feed-in (purple), without any corresponding generation (blue), which seemed a little odd. 

![image](https://user-images.githubusercontent.com/568104/227764863-5d958305-3775-4074-83f6-007d50b2996a.png)

So I went looking into the logs. The only actual error around that time (I don't have debug logs enabled) is:

```
2023-03-26 06:34:00.495 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 543, in async_update_ha_state
    self._async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 597, in _async_write_ha_state
    state = self._stringify_state(available)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 562, in _stringify_state
    if (state := self.state) is None:
  File "/usr/src/homeassistant/homeassistant/components/sensor/__init__.py", line 855, in state
    value = self.native_value
  File "/config/custom_components/foxess_modbus/entities/modbus_charge_period_sensors.py", line 74, in native_value
    value = time(hour=(value & 0xFF00) >> 8, minute=value & 0xFF)
ValueError: minute must be in 0..59
2023-03-26 06:35:10.243 INFO (Recorder) [homeassistant.components.sensor.recorder] Detected new cycle for sensor.grid_consumption_sum_today, value dropped from 1.1 to 2023-03-26T05:34:00.494176+00:00, triggered by state with last_updated set to 0.0
```

The first oddity there is of course that we managesd to get a number of minutes > 0 on the start/end of a charge period (which should be set to midnight I should add, since they're both disabled).

The second is that HA seems to think that grid_consumption_today fell from 1.1 to... a date? If I go looking, it fell to 0.0 for about an hour: 

![image](https://user-images.githubusercontent.com/568104/227765280-171a62fd-56d4-4c90-b540-934542e362c4.png)

It looks like charge period 2 was briefly enabled:

![image](https://user-images.githubusercontent.com/568104/227765374-fccf5e8f-4c2a-41d1-b599-121129e80c4e.png)

With a start time of 00:15 and an end time of midnight:

[...]

PR #86, opened 2023-03-26
Handle incorrect charge period start/end times better

canton7:
Report Unavailable if the value we read doesn't make sense, log the invalid value, and don't update _last_enabled_value.

Relates to: #85


canton7:
@nathanmarlor Hold off on releasing this for a bit -- spotted an issue (by bad!).

I'll push as soon as github fixes their infrastructure...

PR #91, opened 2023-03-28
Added validation for AUX inverter types

nathanmarlor:


custom_components/foxess_modbus/entities/modbus_sensor.py
```diff
@@ -60,11 +63,19 @@ def __init__(
...
         if value is not None:
             if self.entity_description.scale is not None:
                 value = value * self.entity_description.scale
             if self.entity_description.post_process is not None:
-                return self.entity_description.post_process(value)
+                value = self.entity_description.post_process(value)
+
+            rules = self.entity_description.validate
+            if not self._validate(rules, value):
+                _LOGGER.warning(
```
canton7, line 75: Can we log the original value read from the register, as well as the address? Just in case there's a problem with our conversion.

custom_components/foxess_modbus/entities/modbus_sensor.py
```diff
@@ -60,11 +63,19 @@ def __init__(
...
+                    "Value (%s) failed validation against rules (%s)", value, rules
```
canton7, line 76: Might be good to say which rule failed?

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
```diff
@@ -22,6 +25,7 @@
         state_class=SensorStateClass.MEASUREMENT,
         native_unit_of_measurement="V",
         scale=0.1,
+        validate=[Range(0, 1000)],
```
canton7, line 28: Ranges are exclusive, but 0 is a valid value. I'd be tempted to make all of the ranges inclusive

custom_components/foxess_modbus/entities/modbus_sensor.py
```diff
@@ -81,3 +92,7 @@ def state_class(self) -> SensorStateClass:
     @property
     def should_poll(self) -> bool:
         return False
+
+    def _validate(self, rules, value) -> bool:
+        """Validate against a set of rules"""
+        return all(rule.validate(value) for rule in rules)
```
canton7, line 98: `rules` can be `None`, according to its type annotation?

canton7 (commented):
Looks good! :+1:

I was thinking of doing something similar with e.g. `expected_min` and `expected_max` on the entity description, but I think this is probably neater.

custom_components/foxess_modbus/entities/modbus_sensor.py
nathanmarlor: Thought i'd set it to `field(default_factory=list)

canton7: Right, but someone can assign `validate = None`, which is valid according to the type hint, but that will make this function crash. mypi would be moaning at you here :)

canton7: (I'd just remove `| None` from the type hint if you don't intend to allow `None`)

nathanmarlor:
I think that's everything - thanks, some good spots in there :)

Now just to go through and define validation rules for 100+ sensors :laughing:

canton7:
Awesome, looks good to me!

nathanmarlor:
@FozzieUK - it'd be great if you could sense check the validation rules I've applied here please?!

custom_components/foxess_modbus/entities/modbus_entity_mixin.py
```diff
@@ -74,3 +67,29 @@ def update_callback(self, changed_addresses: set[int]) -> None:
...
+        """Get unique ID"""
+        friendly_name = self._inv_details[FRIENDLY_NAME]
+        if friendly_name != "":
+            return f"{friendly_name}_{self.entity_description.key}"
+        else:
+            return f"{self.entity_description.key}"
+
+    def _validate(self, rules: list[BaseValidator], processed, original=None) -> bool:
+        """Validate against a set of rules"""
+        original = original if not None else processed
```
canton7, line 81: `original = original if original is not None else processed`.

Currently this is the same as `original = original`, since `not None` is always True.

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
```diff
@@ -39,7 +43,7 @@
         device_class=SensorDeviceClass.POWER,
         state_class=SensorStateClass.MEASUREMENT,
         native_unit_of_measurement="kW",
-        scale=0.001,
```
canton7, line 42: We lost the scale?

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
```diff
@@ -167,6 +184,8 @@
         state_class=SensorStateClass.TOTAL,
         native_unit_of_measurement="kWh",
         scale=0.1,
+        # currently overflows
```
canton7, line 187: I'd expect this might be negative if the inverter took more energy from the AC side than it output?

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
```diff
@@ -176,6 +195,7 @@
         state_class=SensorStateClass.TOTAL_INCREASING,
         native_unit_of_measurement="kWh",
         scale=0.1,
+        validate=[Range(0, 100)],
```
canton7, line 198: Ditto

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
```diff
@@ -429,12 +476,15 @@
...
         scale=0.1,
+        # currently overflows
+        # validate=[Range(0, 100)]
     ),
     ModbusSelectDescription(
         key="work_mode",
         address=41000,
         name="Work Mode",
         options_map={0: "Self Use", 1: "Feed-in First", 2: "Back-up"},
+        validate=[Range(0, 2)],
```
canton7, line 487: I don't think this is necessary? See comment above.

custom_components/foxess_modbus/entities/modbus_select.py
```diff
@@ -57,6 +60,10 @@ def __init__(
     @property
     def current_option(self) -> str | None:
         value = self._controller.read(self.entity_description.address)
+        if value is None:
+            return None
+        if not self._validate(self.entity_description.validate, value):
+            return None
         return self.entity_description.options_map.get(value)
```
canton7, line 67: This line is effectively checking whether the value is one that's valid, as it returns `None` if the key isn't found in `options_map`. Maybe we should add some logging around this?

custom_components/foxess_modbus/entities/modbus_entity_mixin.py
nathanmarlor: :)

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
nathanmarlor: Unsure, have left it commented out for now.

custom_components/foxess_modbus/entities/modbus_select.py
nathanmarlor: I went for the belt & braces approach and ensured consistency of validation across the code base - we could end up doing something different with the options map in the future (which doesn't return None by default) and/or need additional validation rules for other selects?

canton7: `options_map.get` does return None if the key isn't found -- that's why I used `.get`. Agreed we might want additional validation rules in the future (although I can't think what they might be at the moment!)

I just dislike having duplicate knowledge spread around, as someone needs to manually keep them in sync. We already need knowledge of which values are valid so that we can map them onto strings, IMO duplicating that knowledge into a `Range` is just increasing the maintenance burden.

custom_components/foxess_modbus/entities/modbus_select.py
```diff
@@ -64,7 +64,13 @@ def current_option(self) -> str | None:
...
         if not self._validate(self.entity_description.validate, value):
             return None
-        return self.entity_description.options_map.get(value)
+
+        options = self.entity_description.options_map
+        if value not in options:
+            _LOGGER.warning(
+                "Value (%s) is not valid. Valid values: (%s)", value, self.options
+            )
+        return options.get(value)
```
canton7, line 73: Nit: we can avoid the double-lookup by doing `get` once, then checking whether we got `None`.

Also log the entity id / address maybe?

canton7:
:shipit: from me, perhaps with the LAN validations as well?

nathanmarlor:
I'll merge this then throw up the LAN validations in another PR - suggest we leave this at beta for a while and see if anyone reports anything odd?

nathanmarlor:
Warnings/errors bubble up to here - but it's a bit of a mess...

image

FozzieUK:
I'm running the latest beta, all looks good with the Aux validations so far apart from Grid Consumption Total (register 11082), in my case it's a large negative I assume it's because I make use of low tariff to charge 18kwh of batteries and discharge throughout the day.

image

canton7:
Oh interesting! We might need to tweak that sensor then... How does that relate to your feed-in total sensor?

@nathanmarlor it would be useful to log the entity id I think...

nathanmarlor:
> I'm running the latest beta, all looks good with the Aux validations so far apart from Grid Consumption Total (register 11082), in my case it's a large negative I assume it's because I make use of low tariff to charge 18kwh of batteries and discharge throughout the day.
> 
> image
> 

I knew there'd be on with a huge system which broke things :) I'll fix it later on today.

nathanmarlor:
I had a feeling this might be the case.

nathanmarlor:
Well that was easier than expected - introduced the option to set sensors as signed/unsigned.

image


image

FozzieUK:
Ok, loaded up and looking good - i'm back at grid consumption total of 4686.2 kwh, confirmed at the inverter panel :)

canton7:
Looks like none of them have overflowed 16 bits yet. Can you remember to take a look when one of them passes 6,554 kwh?

FozzieUK:
@canton7 ok, will do :+1:

Now the BMS Watthours is unsigned and is positive I can see the problem, it's scale is 0.1 - it shouldn't be.

I've modified my install with the scale removed and it's climbing upwards in 1 watt increments - i've knocked up a quick and dirty battery_watt_rate sensor which lets me measure the approx battery watt changes and it looks to be correct.

I'll check it over a much longer period and report back to confirm.

image

nathanmarlor:
> Looks like none of them have overflowed 16 bits yet. Can you remember to take a look when one of them passes 6,554 kwh?

It'll be a while but I'm sure it'll be obvious :+1:

FozzieUK:
I know the charge times isn't complete yet but as its validation related thought i'd mention it in case there is something useful here - I noticed these errors last night after changing my charge times. 

The charge times set were Force/Grid on period 1 00:30-04:30, period 2 both off time 00:00-00:00

Data string sent to foxess_modbus write register service:-
Register:  41001 
Friendly name: ""
Data: '1, 30, 1054, 0, 0, 0'

2023-03-31 21:41:54.256 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 546, in async_update_ha_state
self._async_write_ha_state()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 600, in _async_write_ha_state
state = self._stringify_state(available)
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 565, in _stringify_state
if (state := self.state) is None:
File "/usr/src/homeassistant/homeassistant/components/binary_sensor/__init__.py", line 211, in state
if (is_on := self.is_on) is None:
File "/config/custom_components/foxess_modbus/entities/modbus_charge_period_sensors.py", line 203, in is_on
or not self._validate(rules, start_time)
File "/config/custom_components/foxess_modbus/entities/modbus_entity_mixin.py", line 85, in _validate
if not rule.validate(processed):
File "/config/custom_components/foxess_modbus/entities/validation.py", line 59, in validate
hours, minutes = ((data & 0xFF00) >> 8, data & 0xFF)
TypeError: unsupported operand type(s) for &: 'str' and 'int'

2023-03-31 21:41:54.263 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 546, in async_update_ha_state
self._async_write_ha_state()
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 600, in _async_write_ha_state
state = self._stringify_state(available)
File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 565, in _stringify_state
if (state := self.state) is None:
File "/usr/src/homeassistant/homeassistant/components/binary_sensor/__init__.py", line 211, in state
if (is_on := self.is_on) is None:
File "/config/custom_components/foxess_modbus/entities/modbus_charge_period_sensors.py", line 203, in is_on
or not self._validate(rules, start_time)
File "/config/custom_components/foxess_modbus/entities/modbus_entity_mixin.py", line 85, in _validate
if not rule.validate(processed):
File "/config/custom_components/foxess_modbus/entities/validation.py", line 59, in validate
hours, minutes = ((data & 0xFF00) >> 8, data & 0xFF)
TypeError: unsupported operand type(s) for &: 'str' and 'int'

Despite the error the charge times were updated normally 👍

canton7:
Ah. That's a bug, thanks.

PR #94, opened 2023-03-28
Automatically configure energy dashboard through config flow

nathanmarlor:


nathanmarlor:
image

custom_components/foxess_modbus/config_flow.py
```diff
@@ -212,6 +224,35 @@ async def _autodetect_modbus(self, inv_type, host, slave):
             self._errors["base"] = "modbus_error"
         return False, None
 
+    async def _setup_energy_dashboard(self):
+        """Setup Energy Dashboard"""
+        manager = await data.async_get_manager(self.hass)
+        energy_prefs = EnergyPreferencesUpdate(
+            energy_sources=[
+                SolarSourceType(
+                    type="solar", stat_energy_from="sensor.solar_sum_total"
```
canton7, line 233: Doesn't this assume that there's no friendly_name?

canton7, line 233: Might it be worth waiting until we get PV1 and PV2 sums into place?

.vscode/launch.json
```diff
@@ -9,6 +9,7 @@
       "request": "attach",
       "port": 5678,
       "host": "localhost",
+      "justMyCode": false,
```
canton7, line 12: Accidental addition?

custom_components/foxess_modbus/config_flow.py
```diff
@@ -58,7 +66,8 @@ def __init__(self, config=None) -> None:
             {
                 vol.Required(INVERTER_TYPE, default="TCP"): selector(
                     {"select": {"options": ["TCP", "SERIAL"]}}
-                )
+                ),
+                vol.Required(ENERGY_DASHBOARD, default=True): bool,
```
canton7, line 70: Maybe default to false? Nuking the user's current set-up feels a bit dangerous for something which is enabled by default...

canton7 (commented):
I'm surprised it's this easy to be honest!

.vscode/launch.json
nathanmarlor: Perhaps should push it into a different commit to make it obvious, but was needed to understand what was happening in the HA energy validation routines :)

canton7: Yeah I've been enabling it locally as well! Happy to include it if it's deliberate

custom_components/foxess_modbus/config_flow.py
nathanmarlor: But it's so cool to do it automatically... why would anyone not want it?! :laughing:

custom_components/foxess_modbus/config_flow.py
```diff
@@ -212,6 +224,35 @@ async def _autodetect_modbus(self, inv_type, host, slave):
...
+        energy_prefs = EnergyPreferencesUpdate(
+            energy_sources=[
+                SolarSourceType(
+                    type="solar", stat_energy_from="sensor.solar_sum_total"
+                ),
+                GridSourceType(
+                    type="grid",
+                    flow_from=[
+                        FlowFromGridSourceType(
+                            stat_energy_from="sensor.grid_consumption_sum_total"
```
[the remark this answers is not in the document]
canton7: Currently it's also just available for H1 as well -- it's probably on AC1 etc, but I don't have one to test. It might be on LAN as well for all I know...

If we put in a reimann sum for this on the platforms which don't have a register for now, will it break everything if we later remove that and replace it with a proper register later?

nathanmarlor:
I'll update what I can for now, then leave this until we have Reimann sums for PV1/PV2 and grid/feed-in for LAN.

custom_components/foxess_modbus/config_flow.py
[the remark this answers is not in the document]
canton7: Actually, at the least it would either throw away all history or cause a step-change, I think, neither of which is great.

custom_components/foxess_modbus/config_flow.py
canton7: E.g. I've got thing set up slightly differently: I don't have my battery yet so I get nonsense for some of the battery sensors, and don't have that added to the energy dashboard, and I'm still using foxess-ha for some sensors until I can work out how to properly migrate my history over. I also need to remove/readd the integration a fair bit, as config changes etc. It only takes one instance of forgetting to uncheck the checkbox and it's gone and nuked everything...

nathanmarlor:
Single (with name):
Screenshot from 2023-03-30 11-15-14

Single (no name):
Screenshot from 2023-03-30 11-13-15

Multiple (with names):
Screenshot from 2023-03-30 11-12-36

Not sure if the setup is correct for multiple - however this is probably only 5-10% of users, and it's un-selected by default. Any thoughts?

canton7:
Maybe show a message recommending that people check the config themselves after setting it up (can we put in a button / link?). Then people with multiple inverters are promoted to take an interest, rather than just looking at the result, not liking it, and opening an issue?

It's probably easier for the user to remove entities than add them, so I'm in favour of adding evenetjtinf relevant then letting them trim it down if they want.

iancg:
The Solar Panels and Home Battery Storage are correct for dual inverter.  For mine I only put one of the two inverters in the Grid side otherwise (depending on which inverter is shouldering the load) it can double count - but I don't think there is any way you can automatically select which inverter to pick (unless you just pick the first and suggest people put their "outer" inverter in first).

PR #98, opened 2023-03-29
Added integration sensors to replace manually configured yaml

nathanmarlor:


custom_components/foxess_modbus/entities/modbus_integration_sensor.py
```diff
@@ -0,0 +1,109 @@
...
+
+        self._controller = controller
+        self._entry = entry
+        self._inv_details = inv_details
+        self.entity_description = entity_description
+        self.entity_id = "sensor." + self._get_unique_id()
+
+        friendly_name = self._inv_details[FRIENDLY_NAME]
+        if friendly_name != "":
+            source_entity = f"sensor.{friendly_name}_{source_entity.split('.')[1]}"
```
canton7, line 84: Maybe either use `source_entity.split('.')[0]` instead of hard-coding `sensor`, or don't specify `sensor` in `ModbusIntegrationSensorDescription.source_entity` and add here regardless of whether the friendly name is set? Currently we've got a situation where you have to specify `sensor.` in `source_entity`, but things might break if you specify something other than `sensor.`

canton7:
I'm not going to have access to a computer for a few days so I can't give it a test run unfortunately, but it looks like exactly what I was thinking 👍

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
```diff
@@ -46,6 +50,18 @@
         scale=0.001,
         validate=[Range(0, 10000)],
     ),
+    ModbusIntegrationSensorDescription(
+        key="pv1_power_total",
+        address=11002,
```
canton7, line 55: Do we have "power" in the names of other sensors which denote energy? Maybe "pv1_sum_total" to match the solar_sum_total? Or is there a sensor from StealthChestnut's integration we want to copy the name of?

custom_components/foxess_modbus/entities/modbus_integration_sensor.py
```diff
@@ -0,0 +1,109 @@
...
+from .entity_factory import EntityFactory
+
+_LOGGER = logging.getLogger(__name__)
+
+
+@dataclass(kw_only=True)
+class ModbusIntegrationSensorDescription(SensorEntityDescription, EntityFactory):
+    """Custom sensor description"""
+
+    address: int
```
canton7, line 27: I don't think we use the address for anything. Can we remove it (and return an empty list from `addresses`)

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
```diff
@@ -46,6 +50,18 @@
...
+        device_class=SensorDeviceClass.ENERGY,
+        state_class=SensorStateClass.TOTAL,
+        native_unit_of_measurement="kWh",
+        integration_method="left",
+        name="PV1 Power Total",
+        round_digits=2,
+        source_entity="sensor.pv1_power",
```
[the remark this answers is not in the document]
iancg: Will this work where the inverter has a friendly name prefix (which dual inverter setups need)?

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
nathanmarlor: I matched against previous convention - `key="feed_in_sum_total"` StealthChestnuts is `pv1_sum` which isn't descriptive enough IMHO - sum of power? voltage? current?

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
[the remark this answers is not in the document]
canton7: Yes, there's code above which copes with that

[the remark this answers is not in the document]
nathanmarlor: Unfortunately not - the entity id (`sensor.pv1_power`) is the only unique identifier of this sensor, if it's renamed all bets are off!

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
canton7: Hmm. I guess `pv1_energy_total` is strictly correct, but goes against what we have for feed_in etc which don't specify whether it's energy/power (and use "sum" to imply its energy). I guess "sum" on its own could mean charge, which is the integral of current...

nathanmarlor: They were only released under beta so I think there's an opportunity to rename. Suggest:

`feed_in_energy_today`
`feed_in_energy_total`
...
`pv1_energy_total`

canton7: Yeah I like the sound of those. Thanks!

custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py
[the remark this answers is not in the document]
canton7: I guess if we could get a reference to the ModbusSensorEntity instance representing the PV1 sensor, we should be able to query the renamed entity_id from it? Since HA must be able to link the two. Maybe we can event query it for a link between original and renamed entity id somehow, since it must know. I've no idea how to go about it though. I suggest we get this in as is and worry about that later!

custom_components/foxess_modbus/entities/modbus_integration_sensor.py
```diff
@@ -24,7 +24,8 @@
 class ModbusIntegrationSensorDescription(SensorEntityDescription, EntityFactory):
     """Custom sensor description"""
 
-    address: int
+    # Unused for this sensor type
+    address = 0
```
canton7, line 28: I think we can probably remove the field entirely?

nathanmarlor: Unfortunately not...

```
    def update_callback(self, changed_addresses: set[int]) -> None:
        """Schedule a state update."""
        if self.entity_description.address in changed_addresses:
```

canton7: Ah, that might be worth improving at some point.

PR #100, opened 2023-03-29
Added differentiation between signed/unsigned numbers

nathanmarlor:


custom_components/foxess_modbus/entities/modbus_sensor.py
```diff
@@ -26,6 +26,7 @@ class ModbusSensorDescription(SensorEntityDescription, EntityFactory):
     scale: float | None = None
     post_process: Callable[[int], int] | None = None
     validate: list[BaseValidator] = field(default_factory=list)
+    signed: bool | None = True
```
canton7, line 29: Do we want to allow None here?

nathanmarlor: Nope... will fix then I'm done for the day!

PR #111, opened 2023-04-03
Part 2 of properly supporting charge windows

canton7:
Add a new service which can update one of the two charge windows. This ensures that the time period being written is valid:

 * Just consists of hours and minutes
 * In the range 00:00 -> 23:59
 * End is > start
 * Period does not overlap any other periods configured on the inverter

It then gathers together all time periods, and writes them in a single write.

image

The inverter can be specified either using a drop-down in the GUI, or by providing a friendly name. This also updates the existing "write_registers" service to follow the same scheme.

### DIFF
--- a/custom_components/foxess_modbus/entities/modbus_charge_period_sensors.py
+++ b/custom_components/foxess_modbus/entities/modbus_charge_period_sensors.py
@@ -4,7 +4,6 @@ from dataclasses import dataclass
 from dataclasses import field
 from datetime import time
 
-from custom_components.foxess_modbus.entities.validation import BaseValidator
 from homeassistant.components.binary_sensor import BinarySensorDeviceClass
 from homeassistant.components.binary_sensor import BinarySensorEntity
 from homeassistant.components.binary_sensor import BinarySensorEntityDescription
@@ -17,14 +16,25 @@ from homeassistant.helpers.restore_state import RestoredExtraData
 from homeassistant.helpers.restore_state import RestoreEntity
 
 from ..common.entity_controller import EntityController
+from .base_validator import BaseValidator
 from .entity_factory import EntityFactory
 from .modbus_entity_mixin import ModbusEntityMixin
 
 _LOGGER = logging.getLogger(__name__)
 
 
-def _parse_time(value: int) -> tuple[int, int]:
-    return ((value & 0xFF00) >> 8, value & 0xFF)
+def is_time_value_valid(value: int) -> bool:
+    hours, minutes = ((value & 0xFF00) >> 8, value & 0xFF)
+    return 0 <= hours <= 23 and 0 <= minutes <= 59
+
+
+def parse_time_value(value: int) -> time:
+    hours, minutes = ((value & 0xFF00) >> 8, value & 0xFF)
+    return time(hour=hours, minute=minutes)
+
+
+def serialize_time_to_value(time_value: time) -> int:
+    return (time_value.hour << 8) | time_value.minute
 
 
 def _is_force_charge_enabled(
@@ -100,8 +110,7 @@ class ModbusChargePeriodStartEndSensor(ModbusEntityMixin, RestoreEntity, SensorE
             and not _is_force_charge_enabled(value, other_value)
         ):
             value = self._last_enabled_value
-        hours, minutes = _parse_time(value)
-        value = time(hour=hours, minute=minutes)
+        value = parse_time_value(value)
 
         return value
 
--- a/custom_components/foxess_modbus/__init__.py
+++ b/custom_components/foxess_modbus/__init__.py
@@ -98,6 +98,13 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         _WRITE_SCHEMA,
     )
 
+    hass.services.async_register(
+        DOMAIN,
+        "update_charge_period",
+        lambda data: None,
+        _WRITE_SCHEMA,
+    )
+
     hass.data[DOMAIN][entry.entry_id] = {
         INVERTERS: inverter_controller,
     }
--- a/custom_components/foxess_modbus/__init__.py
+++ b/custom_components/foxess_modbus/__init__.py
@@ -7,15 +7,11 @@ https://github.com/nathanmarlor/foxess_modbus
 import asyncio
 import logging
 
-import voluptuous as vol
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import Config
 from homeassistant.core import HomeAssistant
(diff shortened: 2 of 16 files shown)


canton7:
There's a couple of things I need to tidy up:

* Make sure the inverter actually supports time periods. I suspect it might just assert currently.
* Make sure that the pre-existing time period has a value for enable force charge from grid
* Add a service which sets both charge periods together. We might not want to expose this in services.yaml, as I want to pass an array of charge periods and I don't think the UI supports this. This will be used by the custom card in part 3

iancg:
> Make sure that the pre-existing time period has a value for enable force charge from grid

I often have two time periods set where neither has force charge from grid (to give the equivalent functionality of setting the mode to backup which previously wasn't available to me, but also limits the "backup" mode to just a time period).  Setting a time period with no force charge from grid causes the inverter to charge the battery from the solar but not send any to the house.  I use this to balance the two inverters - if house load is low (i.e. so it doesn't need both inverters) then the inverter with the lower battery is given a time period without force charge to top its battery up.

> Add a service which sets both charge periods together. We might not want to expose this in services.yaml, as I want to pass an array of charge periods and I don't think the UI supports this. This will be used by the custom card in part 3

If I've understood correctly, the built in modbus component does this (i.e. support arrays) - 

      service: modbus.write_register
      data:
        hub: FoxESSInverterModbus
        address: 41001
        slave: 246
        value:
          - 1
          - 31
          - 1053
          - 0
          - 0
          - 0

the values are an array in this case.  I hadn't tried putting templates in individual values of the array, but I don't see why this wouldn't work.

canton7:
> I often have two time periods set where neither has force charge from grid (to give the equivalent functionality of setting the mode to backup which previously wasn't available to me, but also limits the "backup" mode to just a time period). Setting a time period with no force charge from grid causes the inverter to charge the battery from the solar but not send any to the house. I use this to balance the two inverters - if house load is low (i.e. so it doesn't need both inverters) then the inverter with the lower battery is given a time period without force charge to top its battery up.

You misunderstood my note I think. When the integration first starts, we take about 10 seconds to fetch the current value of the registers. We need to program both charge periods together, but the service lets you update just one of the charge periods at a time, which means it needs to know what the current value for the other charge period is (so we can write those same values back). But it doesn't know this if it hasn't yet fetched those registers. The PR already checks whether the start/end time have been fetched correctly, but not if enable charge from grid has been fetched. 

> If I've understood correctly, the built in modbus component does this (i.e. support arrays)

I want to pass an array of objects, e.g.

```yaml
data:
  - start:...
    end:... 
   ... 
  - start:... 
    end:... 
   ... 
```

This is for use by the card, not normal users. I'd be very surprised if the HA services UI could represent this - nothing I've seen in the docs indicates that it can!

iancg:
> You misunderstood my note I think.

Yes - I had, thanks for the clarification, that makes sense.

> I want to pass an array of objects

Next level! OK, that also makes sense.

Thanks.

canton7:
There we go!

custom_components/foxess_modbus/entities/modbus_charge_period_sensors.py
```diff
@@ -100,8 +110,7 @@ def native_value(self):
             and not _is_force_charge_enabled(value, other_value)
         ):
             value = self._last_enabled_value
-        hours, minutes = _parse_time(value)
-        value = time(hour=hours, minute=minutes)
+        value = parse_time_value(value)
```
nathanmarlor, line 113: This seems a little back-to-front now? Wouldn't it make more sense to use `validation.py` instance of `Time`?

nathanmarlor (commented):
Apologies it's taken me so long to get caught up - had a ~6 month old SSD fail over the weekend which knocked me back a bit.

canton7:
> In fact, we should probably fix the friendly name bug and add the additional EPS sensors first. I'll try to find time this weekend.

I'm in two minds about those additional sensors -- it's going to be easier to add them in later, than add them now and then change our minds and remove them (or hide them behind some setting).

custom_components/foxess_modbus/entities/modbus_charge_period_sensors.py
canton7: To be honest, I don't particularly like using the validation stuff to check for valid times. The time-related sensors rely on the time being valid to avoid throwing exceptions, so making that validation "configurable" smells.

I tried to keep all of the logic for validating, parsing and serializing time registers inside modbus_charge_period_sensors.py -- validating them requires the same amount of knowledge as parsing/serializing them, and putting that validation logic inside validation.py just duplicates that knowledge around the place.

canton7:
Oh, and I want to move those solar_energy_total sensors into the H1_AC1 set and remove the integration sensors that are currently there before the next release. Moving from an integration sensor to the total value reported from the inverter is going to cause a step jump, which won't be ideal if AC1 users come to rely on it.

At least by making the change now, we'll learn soon enough it those registers aren't supported by the AC1!

nathanmarlor:
I've been in touch with an AC1 user over the last couple of days and he confirmed that the sensors are present :+1:

canton7:
Fab, thanks! That change will conflict with #124, so holding off for a little.

custom_components/foxess_modbus/entities/modbus_charge_period_sensors.py
nathanmarlor: Fair enough - if we were dealing with multiple instances of time from various sources it'd probably be a different story but I can see the logic in keeping everything together. :shipit:

canton7: True, fair enough. In that case we'd probably want to move the time stuff into its own place?

FozzieUK:
@canton7 sorry a bit late but just a quick display thing before it gets released.

The text for the charge windows has ",or just prevent discharge" on the Enable charge from grid - that should just be the text "Whether to charge the battery from the grid"

The Force Charge Window toggle should have the text "Whether to enable this force charge window, or just prevent discharge".

canton7:
@FozzieUK The aim was to indicate what happened if the option was deselected. I'll clarify.

FozzieUK:
@canton7 Ok, thanks - it's one of the lesser understood things about the force charge switches; basically the Enable Force Charge switch on it's own doesn't enable force charge, it just prevents the battery from being discharged during the set charge period, whereas the Enable charge from grid, force charges from grid and it doesn't allow the batteries to discharge when they batteries hit maxsoc.

canton7:
Yes, which is what the message was trying to indicate.

PR #124, opened 2023-04-06
Support entities which span multiple registers

### LINKED ISSUE #117: Register 11085 looks like the upper bits of total_yield_total
https://github.com/StealthChesnut/HA-FoxESS-Modbus/issues/112#issuecomment-1497459684

canton7:
This makes the assumption that grid_consumption_energy_total, feed_in_energy_total, battery_discharge_total, battery_charge_total and solar_energy_total follow the same pattern as total_yield_total.

Fixes #117

### DIFF
--- a/custom_components/foxess_modbus/entities/modbus_sensor.py
+++ b/custom_components/foxess_modbus/entities/modbus_sensor.py
@@ -22,19 +22,25 @@ _LOGGER = logging.getLogger(__name__)
 class ModbusSensorDescription(SensorEntityDescription, EntityFactory):
     """Custom sensor description"""
 
-    address: int
+    # Array of registers which this value is split over, from lower-order bits to higher-order bits
+    addresses: list[int]
+    _addresses: list[int] = field(init=False, repr=False)
     scale: float | None = None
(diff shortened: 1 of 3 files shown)


custom_components/foxess_modbus/entities/modbus_sensor.py
```diff
@@ -63,11 +69,19 @@ def __init__(
     @property
     def native_value(self):
         """Return the value reported by the sensor."""
-        value = original = self._controller.read(self.entity_description.address)
-        if value is None:
-            return None
+        original = 0
```
nathanmarlor, line 72: Should we be returning 0 for sensors we can't read? Won't this also set all sensors to 0 on start-up and mess with the history?

canton7: If any of the reads return None, we still return None (line 76).

This just sets `original` as a clean state, onto which the bits from the individual register reads are OR'ed

nathanmarlor: Missed that, thanks.

Issue #135, opened 2023-04-09
Tweak wording on charge service


canton7:
@canton7 sorry a bit late but just a quick display thing before it gets released.

The text for the charge windows has ",or just prevent discharge" on the Enable charge from grid - that should just be the text "Whether to charge the battery from the grid"

The Force Charge Window toggle should have the text "Whether to enable this force charge window, or just prevent discharge".

_Originally posted by @FozzieUK in https://github.com/nathanmarlor/foxess_modbus/issues/111#issuecomment-1501084917_

PR #136, opened 2023-04-09
Tweak the wording for 'force charge from grid'

canton7:
Fixes: #135


canton7:
@FozzieUK is that clear?

FozzieUK:
Yes absolutely, thanks 😊

Issue #122, opened 2023-04-05
If a poll is aborted partway, sensors are never notified of reads which were successful


canton7:
When we do a poll, we read batches of 8 registers and gather up those that have changed, then notify all sensors of the registers which have changed once the all reads are complete.

However, if one of more earlier reads succeed but a later read fails, then we'll change some of the values in `_data`, but won't notify any sensors that they've changed.

One approach would be to abort the poll if any read fails, but notify all sensors of the reads which *were* successful before we aborted. There is however a risk here that we read only part of an internally-consistent set of registers (such as the time period config). (I know there's a chance that a change can happen part-way through a poll normally, but we can normally be confident that another poll will happen again soon. If a read fails, because e.g. communication failed, we might not be able to do another read for a long time, and our record of the inverter's state will stay inconsistent for longer).

Another option is to put all reads in a temporary dict, and then commit that to `_data` and notify all sensors at the same time once all reads have succeeded.

canton7:
Thinking more, we should probably start marking things as None if there's a break in communication. Marking everything as None of any read fails feels perhaps a little draconion, but maybe do it if we fail to complete a full set of reads after a couple of attempts?

canton7:
Yeah, but a single failed poll isn't unheard-of, particularly with the W610.

My current thinking is that we should probably mark everything as None if more than say 5 or 10 polls fail in a row

canton7:
Fixed in #1377

PR #137, opened 2023-04-09
Handle polls which are aborted partway through

canton7:
Previously, if a later read failed, we would records new values for earlier reads but wouldn't notify the sensors which depended on those reads.

This change ensures that we perform all of the reads first, and only if all succeed do we write those to _data and notify the sensors.

Relates-to: #122

### DIFF
--- a/custom_components/foxess_modbus/common/entity_controller.py
+++ b/custom_components/foxess_modbus/common/entity_controller.py
@@ -6,20 +6,46 @@ from abc import abstractmethod
 _LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


canton7:
I haven't had a chance to test this yet -- might have the opportunity tonight. If you get there first, feel free to un-draft!

canton7:
This also doesn't fully fix #122, as we don't handle marking sensors unavailable if there's a persistent communication failure.

canton7:
Right, that marks entities as Unavailable if we lose the connection to the inverter

Issue #106, opened 2023-04-02
Friendly name setting 


nathanmarlor:
Friendly names aren't escaped (i.e. ' ' -> _) so result in erroneous sensors.

canton7:
Would it be worth creating both a friendly name and friendly id in config? That let's us e.g. change how it is escaped over time, without breaking people.

(It also lets me have a prefix to all the sensor ids, to keep things neat, without having a friendly name appear at the end of every sensor name)

iancg:
You need the friendly name in the sensor description if you have two inverters otherwise the graphs just show the same sensor names twice and you can't tell which is which (it was me that asked for that to be added).

canton7:
Right, but I've only got 1 inverter. The aim, which I think other people are doing as well, is to put a prefix onto all of the entity ids (which we don't have my default, to be compatible with StealthChestnut's intigration) - this makes it easier to e.g. use globs to slurp everything into influxdb, rather than having to list every sensor individually.

That isn't what this ossie is about, of course. But I saw the opportunity to kill two birds with one stone here.

canton7:
We also need to make sure that two friendly names don't escape to the same friendly id. That's probably also easier to sort out if we generate both at config time, rather than having to sort it out on the fly each time.

iancg:
It's a good point.  Without wanting to over-complicate the config, I wonder if we need a separate "entity_id prefix" from "Friendly name prefix or suffix"?

With StealthChestnut's integration I'd actually prefixed my Friendly name with the inverter name and the suffix was @nathanmarlor 's alternative implementation (but since the entity id's were prefixed, it mattered not).  I'd mainly used prefix as some of the entity_id's are generated from the escaped name in StealthChestnut's one in some cases - e.g. 

   - name: "Inner-PV Energy Daily"

and I didn't want a mix of prefix and suffix in the entity_id's.

screenagerbe:
It's something I'm struggling with too...

Got StealthChestnut's integration up and running in my PROD environment, but I kind of modified it a lot. I've renamed all name's (added `FoxESS` as prefix) and unique_id's (added `foxess_` as prefix) entries in configuration.yaml, modbusLAN.yaml and all the extra entities which are listed in sensors.yaml and templates.yaml. 
Main reason: The FoxESS installation is not the only monitoring tool in my fuse box and I wanted them as recognizable as possible on all fronts.
And in the near future there will also be data send from a P1-port to HA.

[...]

PR #138, opened 2023-04-09
Ensure friendly names are valid in entity ids

canton7:
I refactored the config flow to use exceptions for error handling, as the global _errors state which is mutated in various places and persists between pages was confusing me, and I wasn't confident in being able to add in an error for the friendly_name field without breaking something! Happy to revert and figure it out if you feel strongly about it :)

I haven't tried to do anything with separating out friendly_name and friendly_id, or any of the other stuff discussed in #106. This is just about making sure that friendly_name matches `\w+`, which I think is what HA uses for entity IDs.

I've tried to test every route through the flow, but I don't have a serial adapter so couldn't test that side to completion. It would be good if you could give it a test as well.

Relates to: #106

### DIFF
--- a/custom_components/foxess_modbus/config_flow.py
+++ b/custom_components/foxess_modbus/config_flow.py
@@ -186,7 +186,7 @@ class ModbusFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
         if host in self._data[inv_type]:
             # TODO: Shouldn't we be checking for duplicate friendly names across all hosts?
             if friendly_name in self._data[inv_type][host]:
-                raise ValidationFailedExeption({"base": "modbus_duplicate"})
+                raise ValidationFailedException({"base": "modbus_duplicate"})
 
     async def async_add_inverter(self, inv_type, host, inverter):
         """Handle a flow initialized by the user."""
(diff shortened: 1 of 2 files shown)


custom_components/foxess_modbus/config_flow.py
```diff
@@ -108,116 +111,105 @@ def __init__(self, config=None) -> None:
...
-            errors=self._errors,
+        return await self._with_default_form(
+            user_input, "energy", self._energy_dash, body
         )
 
-    def detect_duplicate(self, inv_type, host, friendly_name):
+    def ensure_not_duplicate(self, inv_type, host, friendly_name):
         """Detect duplicates"""
         if host in self._data[inv_type]:
+            # TODO: Shouldn't we be checking for duplicate friendly names across all hosts?
```
canton7, line 187: This looked wrong -- we allow duplicate friendly names so long as they are used with different IPs, and duplicate IPs so long as they have different friendly_names. Is that correct? I would have thought that friendly names should be unique across all connection types and addresses.

canton7: I'm addressing this in the follow-up PR which reworks the config flow

custom_components/foxess_modbus/config_flow.py
```diff
@@ -327,6 +343,9 @@ async def async_step_init(self, user_input=None):
             }
         )
 
-        return self.async_show_form(
-            step_id="init", data_schema=options_schema, errors=self._errors
-        )
+        return self.async_show_form(step_id="init", data_schema=options_schema)
+
+
+class ValidationFailedExeption(Exception):
```
nathanmarlor, line 349: Small typo -> Exeption : Exception

canton7: Good spot, thanks!

canton7:
It's conflicting as well... Will fix

canton7:
Done!

canton7:
I don't think we want to close #106 when this goes in -- there's still an open question about specifying a name to be appended to entity names which isn't restricted to snake_case

Issue #141, opened 2023-04-12
If poll speed is too high, previous and next reads overlap


canton7:
If the poll speed it too high, the next read starts to happen while the previous read is still ongoing, and the two overlap. This builds up until everything grinds to a halt.

We should either start the next read a set amount of time after the previous read finishes, or abort a read if the previous read is still ongoing.

PR #148, opened 2023-04-13
Abort a refresh if the previous refresh is still in progress

canton7:
Also improve the logging in this area. With multiple inverters it was hard to tell which inverter a log message related to sometimes

Fixes: #141


nathanmarlor:
@canton7 can you take a look at the merge conflicts please?

canton7:
That was painful... There we go!

PR #153, opened 2023-04-15
Fix issues around modbus TCP connections/reconnections

canton7:
Disable Nagle's algorithm after a reconnection.  The socket is recreated on reconnection, and we weren't re-disabling Nagle's.

Remove ModbusClient.connect. It turns out that pymodbus automatically connects the first time that you try and read/write something anyway. I think that the initial first auto_connect connection was messing up my W610 as well, as I often got a failed read on the first read attempt, and would have to wait for the automatic reconnection to happen. Removing the auto_connect seems to fix this.

### DIFF
--- a/custom_components/foxess_modbus/config_flow.py
+++ b/custom_components/foxess_modbus/config_flow.py
@@ -228,7 +228,7 @@ class ModbusFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
                 params.update({"host": host.split(":")[0], "port": host.split(":")[1]})
             else:
                 params.update({"port": host, "baudrate": 9600})
-            client = ModbusClient(self.hass, params, False)
+            client = ModbusClient(self.hass, params)
             return (True, await ModbusController.autodetect(client, slave))
         except UnsupportedInverterException as ex:
             _LOGGER.warning(f"{ex}")
--- a/custom_components/foxess_modbus/modbus_client.py
+++ b/custom_components/foxess_modbus/modbus_client.py
@@ -14,6 +14,20 @@ from .const import TCP
 _LOGGER = logging.getLogger(__name__)
 
 
+class CustomModbusTcpClient(ModbusTcpClient):
+    def __init__(self, **kwargs: any) -> None:
+        super().__init__(**kwargs)
+
+    def connect(self) -> bool:
+        was_connected = self.socket is not None
+        is_connected = super().connect()
+        # pymodbus doesn't disable Nagle's algorithm. This slows down reads quite substantially as the
+        # TCP stack waits to see if we're going to send anything else. Disable it ourselves.
+        if not was_connected and is_connected:
+            self.socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, True)
+        return is_connected
+
+
 class ModbusClient:
     """Modbus"""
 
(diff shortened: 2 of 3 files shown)


nathanmarlor:
Have you tested reconnections - i.e. disconnecting the adapter/485 for ~5 mins then reconnecting?

The native HA modbus integration also maintains the connection in a background thread -> https://github.com/home-assistant/core/blob/dev/homeassistant/components/modbus/modbus.py#L339

canton7:
I tested disconnecting for long enough that connection was dropped, then watched it reconnect.

I'll take another look at the modbus intigration again tomorrow, try and work out why it's doing what it's doing. It might be related to the need to add a delay after a connection before using it? I don't immediately see anything in there to handle a reconnection as well, although I am on mobile.

I wanted to spend some time looking into what actually happens if you tell the modbus intigration to poll a sensor with the W610, which we don't think can handle that - whether it does something different which makes it all work, or whether it just silently falls over. I'll debug into (re-) connections as part of that.

I do suspect that we are telling pymodbus to connect on one thread, then doing a read which is triggering a connection on another thread, and the two are interfering somehow. I very often get an error saying the read failed with 0 bytes read some minescule amount of time into the transaction when HA first starts, and this PR appears to resolve it.

canton7:
I did some archaeology... That `connect()` call was first introduced here, as part of the move to async: https://github.com/home-assistant/core/pull/34043 . It just slipped in - noone commented on it, and it wasn't justified. After that it went through various changes as other things were refactored and improved, but it seems that the assumption was that it was always needed (and maybe it is for the async client, I'm not sure).

canton7:
This is the reconnect call, FWIW: https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/client/base.py#L180

nathanmarlor:
Good digging! I'll take a deeper look and take it for a test drive when I'm back on the laptop but seems to make sense.

I think the connect can also be done as part of the class initialisation (sync/async) -> https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/client/base.py#L346

canton7:
The docs for pymodbus say that it'll automatically reconnect *after a successful initial connection*. Well 1) that doesn't appear to be accurate - you don't actually need the initial connection, and 2) neither us nor the HA integration make sure that the initial connection is successful - we just give it a punt, and if it faisl fall back to the built in reconnection logic.

Also, the HA integration calls connect with the lock held, and we don't. I suspect that's important and avoids the issue I've been seeing, but I'll need to do some more tests with some extra logging to be sure.

canton7:
Hmm OK I added some logging to `connect`, and it doesn't look like the two calls (one during setup, one when we read the first register) overlap in practice (that's just because there's a 10s gap between the initial connection and the first poll). But I still get:

> 2023-04-16 10:27:55.314 DEBUG (MainThread) [custom_components.foxess_modbus.modbus_controller] Modbus exception when polling - Modbus Error: [Connection] ModbusTcpClient(192.168.10.8:502): Connection unexpectedly closed 0.00450444221496582 seconds into read of 8 bytes without response from unit before it closed connection

> Also, the HA integration calls connect with the lock held, and we don't. I suspect that's important and avoids the issue I've been seeing, but I'll need to do some more tests with some extra logging to be sure.

I'm wrong about that. We call `_async_pymodbus_call`, which does take the lock.

However, the error above does seem to stop appearing when I remove the auto_connect logic -- I get the error about 1 time in 2 with the auto_connect present, and I've yet to see it with it absent.

So, I'm more confused this morning than I was last night -- I'm not sure what's going on.

canton7:
Thinking today, I wonder whether the W610 simply didn't like the large gap between the first connection and the first request.

Now that we've reduced that time, however, we might be affecting direct LAN connections. There's talk that people using HA-foxess-modbus needed a `delay: 1` when using a direct LAN connection, and we now don't have any delay here.

PR #155, opened 2023-04-16
Rewrite config flow to add adapter selection

canton7:
I'm submitting this at draft for now, to get some initial feedback. It's based on #138, so I'll rebase it when that gets in. I wouldn't recommend getting this in the next release either, let's give it a bit of time in beta.

At a high level, this rewrites the config flow and configuration format to support letting the user select their exact adapter.

The first step in the config flow is:

"Direct LAN connection" takes you straight through to the IP/port selection (we'll see that in a minute). The other two options give you a list of suitable adapters:

and:

There are several variations of the page you see after selecting one. First, a serial one:

That's fairly unchanged, except that we've moved the friendly name setting to the next page (more on that later), and the option to add another inverter is now slightly later in the flow as well.

Note the link to the wiki with instructions on how to set it up too!

The ethernet one comes in several variations. If it's a direct connection, we just prompt for hostname and slave:

For the other ones, if it supports both TCP and UDP (yes, we now support UDP!), we give them a selector at the top:

And if we want to nudge them in a particular direction, there's a bit of text under the protocol selector:

Whichever route you take, the next option is for the friendly name:

I moved this onto its own page partly to reduce the duplicate validation logic, partly so that the host and friendly name can be validated separately (one concern at a time), and partly to give room for extra field such as name prefix/suffix in the future. It also gives us a chance to properly explain what this field is talking about.

After that we have the option to add another inverter, pulled out to make it a bit more obvious:

And finally the energy dashboard:

------------

The configure menu has also changed a bit. Settings are now per-inverter: if you have multiple inverters configured, you first get to choose which one:

If you only have a single inverter, you don't get this option -- you just go straight to the next page:

This is pretty much the same, except that you can leave the fields blank to use the defaults for that adapter.

--------

~~If you have an inverter configured when you upgrade, we handle migrations. If you're using a serial or ethernet adapter, you'll get prompted to reconfigure:~~

~~Things will keep working in the meantime: we assume they're using the W610 until they tell us otherwise.~~ 

~~They then get prompted to select the adapter for each:~~

(Removed due to probably being too confusing)

----------------

max_read and poll_rate are now part of the adapter configuration, and they aren't stored in the `ConfigEntry.data`. This means that we can change them in future if we realise that a particular adapter needs special handling in some way (that's why I'm trying to gather a list of all adapters now, even if the settings for most are the same -- it lets us change the settings for users with that adapter later). 

Configuration options (`ConfigEntry.options)` are then overlayed on top, so they override whatever the defaults for that adapter are. We no longer duplicate `ConfigEntry.data` into `ConfigEntry.options` -- this lets us migrate them more easily in the future, and it lets the user revert individual settings back to the adapter defaults.

(Configuration options are also migrated currently, with the user's max_read and poll_rate being applied to all of their configured inverters).

I've tweaked the configuration format to be easier to work with: it used to be quite nested, which made it hard to do things like change the settings for an individual inverter from the config flow. It's now more flat -- just a dict of inverter id -> settings dict. On that note I've given each inverter a GUID ID. This is used to correlate settings across `ConfigEntry.data` and `ConfigEntry.options`, and it can stay the same even if we let the user change the friendly name in the future.

### DIFF
--- a/custom_components/foxess_modbus/__init__.py
+++ b/custom_components/foxess_modbus/__init__.py
@@ -75,7 +75,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     # {(modbus_type, host): client}
     clients: dict[tuple[str, str], ModbusClient] = {}
     for inverter_id, inverter in entry.data[INVERTERS].items():
-        options = entry.options.get(INVERTERS, {}).get(inverter_id)
+        # Remember that there might not be any options
+        options = entry.options.get(INVERTERS, {}).get(inverter_id, {})
 
         # Pick the adapter out of the user options if it's there
         adapter_id = options.get(ADAPTER_ID, inverter[ADAPTER_ID])
--- a/custom_components/foxess_modbus/__init__.py
+++ b/custom_components/foxess_modbus/__init__.py
@@ -18,6 +18,8 @@ from .const import MODBUS_SLAVE
 from .const import MODBUS_TYPE
 from .const import PLATFORMS
 from .const import POLL_RATE
+from .const import HOST
+from .const import FRIENDLY_NAME
 from .const import SERIAL
 from .const import STARTUP_MESSAGE
 from .const import TCP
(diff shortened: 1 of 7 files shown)


canton7:
Just seen this...

```
2023-04-17 10:39:26.507 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry FoxESS - Modbus for sensor
Traceback (most recent call last):
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/config_entries.py", line 383, in async_setup
    result = await component.async_setup_entry(hass, self)
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/components/sensor/__init__.py", line 131, in async_setup_entry
    return await component.async_setup_entry(entry)
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/helpers/entity_component.py", line 179, in async_setup_entry
    raise ValueError("Config entry has already been setup!")
ValueError: Config entry has already been setup!
2023-04-17 10:39:26.509 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry FoxESS - Modbus for binary_sensor
Traceback (most recent call last):
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/config_entries.py", line 383, in async_setup
    result = await component.async_setup_entry(hass, self)
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/components/binary_sensor/__init__.py", line 169, in async_setup_entry
    return await component.async_setup_entry(entry)
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/helpers/entity_component.py", line 179, in async_setup_entry
    raise ValueError("Config entry has already been setup!")
ValueError: Config entry has already been setup!
2023-04-17 10:39:26.510 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry FoxESS - Modbus for select
Traceback (most recent call last):
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/config_entries.py", line 383, in async_setup
    result = await component.async_setup_entry(hass, self)
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/components/select/__init__.py", line 112, in async_setup_entry
    return await component.async_setup_entry(entry)
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/helpers/entity_component.py", line 179, in async_setup_entry
    raise ValueError("Config entry has already been setup!")
ValueError: Config entry has already been setup!
2023-04-17 10:39:26.511 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry FoxESS - Modbus for number
Traceback (most recent call last):
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/config_entries.py", line 383, in async_setup
    result = await component.async_setup_entry(hass, self)
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/components/number/__init__.py", line 114, in async_setup_entry
    return await component.async_setup_entry(entry)
  File "/usr/local/python/lib/python3.10/site-packages/homeassistant/helpers/entity_component.py", line 179, in async_setup_entry
    raise ValueError("Config entry has already been setup!")
ValueError: Config entry has already been setup!
```

canton7:
I don't think HA is expecting us to set up any entities if we throw a ConfigEntryAuthFailed -- it seems to leave a controller running in the background, which either get forgotten, or it's reloaded but then it goes and creates another controller on top

canton7:
I think this is probably ready for the first round of review @nathanmarlor

canton7:
I'm undecided on whether I like the whole "RECONFIGURE" bit... It might be less disruptive to just assume that people upgrading are using the W610 (which will give them less-than-perfect performance, but they probably won't notice), and if they care they can always remove and re-add...

canton7:
Looking at the facebook group I'm inclined to remove the "RECONFIGURE" button -- I think too many people are just going to get confused by it

PR #157, opened 2023-04-17
Allow HA to manage unique IDs

nathanmarlor:


nathanmarlor:
image

canton7:
Also from your screenshot this is now adding a suffix to the entity ID?

canton7:
The place I want to get to is the user specifies a friendly name, and we generate a unique entity ID prefix from that (letting the  user customise it if they want?) and store both in config. https://github.com/nathanmarlor/foxess_modbus/issues/106#issuecomment-1493253114. Then we can go back to the user specifying a free-text name, but we're still in control of the ID.

That's why I didn't think #106 was ready to close yet. #138 was just a stop-gap.

nathanmarlor:
Ah, it's reverting back to the entity name (and stripping chars). I'll have another look at this.

canton7:
I don't think there's any way for HA to do this automatically -- you simply can't get around the fact that two friendly names might be escaped to the same thing by HA, yet we need them to be unique *and we need to know which one maps to which friendly name*.

nathanmarlor:
> I don't think there's any way for HA to do this automatically -- you simply can't get around the fact that two friendly names might be escaped to the same thing by HA, yet we need them to be unique _and we need to know which one maps to which friendly name_.

Agreed. Looking through the HA docs they essentially slugify the entity name with retries (appending _1, _2 etc.) on each time. I've mirrored this but I can't decide if it's better or worse :)

image

image

canton7:
That doesn't change to lower case, and I'm not sure what else it might be missing?

It's also bad UX, if we tell them they've for a duplicate friendly name and they entered two different things (which happened to slugify to the same thing)

canton7:
Some people care about what their entity id is, so I'm not sure that springing it on them after they've finished the config flow is nice?

If we had a powerful enough config flow UI I'd let them enter the name, the generate the entity id on the fly and let them edit it. Maybe we could do the same by having two steps? Although this is starting to conflict with my rework of the config flow in that other PR.

iancg:
Presumably the HA API must have a way of finding an entity_id from Friendly name (and potentially return more than one) - can you use that when setting up the power dashboard (i.e. build friendly name, then lookup entity_id from that rather than try and deduce entity_id)?

canton7:
No, HA has no concept of a friendly name.

canton7:
How about something like this?

image

So if you leave it blank, we'll try and generate one ourselves by slugifying the friendly name. If we fail...

image

canton7:
Superseded by #194

PR #167, opened 2023-04-19
Add H3 support

canton7:
First off, it's probably worth doing this one commit by commit... It's pretty large in parts! Sorry.

This comes in a few stages (see the commit messages for more details)

1. Move each entity's address from the EntityDescription to the entity itself. The ModbusController then gets the list of addresses to read from the set of entities which have registered themselves on it
2. Instead of putting one entity address on each EntityDescription, we put a little mapping from inverter model and connection type to address. This means we can ask the EntityDescription whether it supports a particular model and connection type, and it's able to create an entity which has the appropriate address on it. This means we can share entity descriptions across inverter types and connection types
3. It turns out that all of the register definitions for the H3 are for holding registers. Change from using connection type in point 2, to using register type. We no longer say that e.g. all AUX connections use input registers
4. Add the H3 registers.

It turns out that whoever wrote the H3 YAML file managed to identify some registers that we didn't, so copy those across for all register types:

     - input_energy_total
     - input_energy_today
     - load_energy_total
     - load_energy_today
    
(load_energy_total duplicates load_power_total, and it's got the consistent name, so remove load_power_total)


canton7:
This probably wants double-checking against https://github.com/StealthChesnut/HA-FoxESS-Modbus/issues/95#issuecomment-1516366152

canton7:
Argh, looks like the formatter version used by your action is different to the one I have installed...

canton7:
I also want to disable the EPS sensors by default

canton7:
I think this is now pretty merge there... I'll merge tomorrow

benbeton:
thank you for the first beta integration, when adding it to HA i'm receiving the following Info: 
image

H3 Inverter - USR-610

canton7:
Thanks, moved to #197. Would you mind joining me there?

Issue #110, opened 2023-04-03
BMS Watthours sensor


FozzieUK:
Just out of interest i've been getting to the bottom of 'what' the watthours sensor actually is.

I created a watthrate sensor to track it and confirm its value, I had to remove the scale to bring it in line.

![wh3](https://user-images.githubusercontent.com/113460294/229450376-694f4b9e-083b-45dd-aeef-8d27c8650611.jpg)

The only thing I wasn't sure of, was whether the counter was 16 or a 32 bit register - so tracked it for the last few, it eventually got to 65535 and rolled to 0 without any adjacent register incrementing, confirming it is 16 bit, and simply a counter used internally by the BMS as an energy meter to track all battery activity in watts.

So the only changes are the scale 0.1 is incorrect, and should be removed, and the unit of measurement is Wh (not kWh)

https://github.com/nathanmarlor/foxess_modbus/blob/137329b9c8056e0fb73eeaafe5cecfa2cef70abd/custom_components/foxess_modbus/entities/xx1_aux_entity_descriptions.py#L588-L589

canton7:
Nice work! So it's the integral of all (absolute) battery charge and discharge? 

It's a shame that it wraps at 16 bits -- I wonder whether it's worth trying to unwrap that in HA...

Maybe we should /1000 and keep the unit as kWh, for consistency with other energy sensors?)

FozzieUK:
@canton7 yes that's it exactly 👍 

I was really hoping it was going to be an int32 :cry:  - but it's connected to the battery cycle counter, the BMS must do some internal maths on the capacity of batteries installed and increment the cycles so it's probably not worth re-creating a bigger number in HA as you could get there (roughly) by taking your (cycles * known capacity) + watthours.

Actually /1000 would make a lot of sense to keep in kWh (and help the HA database) :smile:

canton7:
To be honest, do we even want this sensor? Is it of any use to anyone?

canton7:
This is now in

PR #168, opened 2023-04-20
Minor fix to BMS watthours to correct scale and unit

FozzieUK:
Just following on from issue #110 this corrects the scale and unit of measurement - a slightly wider decision required as to whether this sensor is needed at all, but at least with these changes it provides a sensor that matches it's name (watthours).


canton7:
I worry that changing the unit might mess up HA -- I've read issues where the oldest/newest unit is applied to all readings from the sensor, ignoring the unit which was present when the reading was taken.

We also record everything else in kWh. Would it be better to rename the sensor?

FozzieUK:
@canton7 Yes agree on the HA stats, they can be fixed in statistics 'Fix Issue' will allow you to update the previous stats or discard.
But i'm thinking like you that there is no value in this sensor and it just adds to the database - i'd rather it was in a correct format Wh or kWh (it's neither at the moment) or simply removed.

canton7:
>  they can be fixed in statistics 'Fix Issue' will allow you to update the previous stats or discard.

Yeah, but only if people know what's happened! And we'd be expecting every user to do this, which is a bit much.

FozzieUK:
@canton7 Lol, i’d be more surprised if they actually noticed with this sensor ;)

Can I leave it with you and @nathanmarlor to decide if we should rescale to kWh or remove it, i’m happy with either.

canton7:
Yeah I don't mind either. Fixing it and then disabling it is a halfway house at least -- people can get it back if they really want.

canton7:
Also, @nathanmarlor seems to be otherwise occupied at the moment and I don't really want to merge anything in myself, so I suspect there's no rush here. Hopefully he'll be back next week when schools are back.

FozzieUK:
@canton7 yeah, I did think about this, my inclination is to leave the entity ID as it is, it still is a watthours sensor but it reports it's value in kwh (despite previously being the wrong scale). I fixed my install to report watthours only so I could compare it with a self generated wH sensor but it's not a total or total_increasing and resets to zero every few days so i'm not concerned about the history jumping.
Yep no worries there's absolutely no rush at all for this, i've been away myself for a few days and just managed to update to the latest release, I noticed when checking everything that the watthours sensor was wrong so thought i'd better do a PR before I forgot :wink: - hopefully Nathan is having a good break 👍

canton7:
Cool, that's a fair point, makes sense. I'm on board!

FozzieUK:
@canton7 ok, no problem - once #167 has been approved/merged, i'll review and update this to match the new format.

canton7:
Cool, or it might make sense to get this in first as it's smaller, in which case that's a reminder to me to update #167

nathanmarlor:
> Also, @nathanmarlor seems to be otherwise occupied at the moment and I don't really want to merge anything in myself, so I suspect there's no rush here. Hopefully he'll be back next week when schools are back.

I've messaged you on FB 🙂

PR #178, opened 2023-04-23
Update README.md

FozzieUK:
Add the extra step after installation via HACS to navigate to Devices & Services, and add the integration


canton7:
Thanks!

Issue #197, opened 2023-04-26
Unknown model connecting to H3


canton7:
Thank you for the first beta integration, when adding it to HA i'm receiving the following Info: 
<img width="592" alt="image" src="https://user-images.githubusercontent.com/1405209/234666150-bbbff5d5-35f5-4034-b45a-fe0359abce9a.png">

H3 Inverter - USR-610

_Originally posted by @benbeton in https://github.com/nathanmarlor/foxess_modbus/issues/167#issuecomment-1523852010_

canton7:
@benbeton can you look in your log (Settings - System - Logs) for warning messages from foxess_modbus? There should be one saying what model the inverter reported itself as.

benbeton:
Log: 
<img width="474" alt="image" src="https://user-images.githubusercontent.com/1405209/234667707-e9441f6f-9e62-4ca9-a534-99280c054229.png">

canton7:
Oh the bugger, it's started it with a space. Right, I'll try and get a fix in later tonight

canton7:
Thanks!

canton7:
Actually, I assume that's a space. Can you just use modpoll to read holding register 30000 for me? (with `-t 4`)

benbeton:
is the something wrong with my cmd? 
<img width="897" alt="image" src="https://user-images.githubusercontent.com/1405209/234682056-cbb457a5-f295-4a3e-a5e5-1042b597e46a.png">

canton7:
I think you're reading too many registers (and some of the later ones are invalid) -- just `-c 5` or so.

canton7:
Try v1.4.1b4

benbeton:
that did the trick, works fine:  I can see all the values incoming
<img width="689" alt="image" src="https://user-images.githubusercontent.com/1405209/234689950-b57b98d0-4a87-4052-9fb1-3c0c2680ed51.png">

thank you, that's great!

canton7:
Awesome, thanks for confirming! Let me know if any of the values don't look quite right

PR #198, opened 2023-04-26
Fix auto-detection of H3 inverters

canton7:
Their model seems to start with a space.

Fixes: #197


canton7:
Just waiting for confirmation from @benbeton that it does start with a space before I merge

canton7:
Actually let's just get it in -- it logs more if it doesn't work, anyway

PR #203, opened 2023-04-27
Don't accidentally save MAX_READ and POLL_RATE to config_entry.data

canton7:
These will have been correctly saved to config_entry.options during the migration to version 2


canton7:
Don't merge. This might actually be #204

Issue #382, opened 2023-07-31
Properly migrate old invalid entity ID prefixes


canton7:
This is confusing the charge period card, see https://github.com/nathanmarlor/foxess_modbus/discussions/380

PR #383, opened 2023-07-31
Migrate old invalid entity ID prefixes

canton7:
This is confusing the charge period card.

This change also opens the door to us letting people reconfigure the friendly name and entity id prefix, as this now won't change the unique id.

Fixes: #382


canton7:
We should make sure that we don't end up with duplicate unique ids. I'm not sure whether it's possible currently...

Issue #480, opened 2023-12-23
 HA 2024.1 introduces a breaking change


canton7:
Looks like we'll need to do something like:

```py
class EntityFactoryMetaclass(type(EntityDescription), type(ABC)):
    pass

class ModbusBinarySensorDescription(BinarySensorEntityDescription, EntityFactory, metaclass=EntityFactoryMetaclass):
    ...
```

Alternatively, it looks like this might also work:

```py
class EntityFactory(ABC, metaclass=EntityFactoryMetaclass):
    ...
```

I'm not sure whethe rthis will work before EntityDescription's metaclass was introduced. Needs testing...

_Originally posted by @canton7 in https://github.com/nathanmarlor/foxess_modbus/discussions/479#discussioncomment-7933836_

FozzieUK:
@canton7 

Antony, my system has been working on the latest version since you modified the code, today I upgraded to the latest beta candidate and the Foxess_modbus integration fails to load with a metaclass warning.

I had another integration (Sun2) which was updated to support the new FrozenOrThawed metaclass and that has survived the upgrade.

I've tried a complete wipe, re-install and rebuilt it from scratch and confirmed it is the latest beta of HA Core which causes the problem.

Sadly it looks as if something wasn't stable on the dev release, i've downgraded back to it and Foxess_modbus works ok again - so seemingly there is still a breaking change in the current beta release candidate of core.

Previous (dev) ha core release - 2024.1.0.dev20231218

Latest (breaking) beta ha core release - 2024.1.0b2

I've included the logs below

```
2023-12-30 09:58:02.803 ERROR (MainThread) [homeassistant.loader] Unexpected exception importing component custom_components.foxess_modbus
Traceback (most recent call last):
File "/usr/src/homeassistant/homeassistant/loader.py", line 822, in get_component
ComponentProtocol, importlib.import_module(self.pkg_path)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/importlib/__init__.py", line 126, in import_module
return _bootstrap._gcd_import(name[level:], package, level)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
[...]

PR #482, opened 2023-12-23
Make dataclasses frozen

canton7:
HA is moving to a model where dataclasses need to be frozen (to aid caching), and will start logging depreciation warnings in 2024.x. Let's get ahead of the game.

See https://github.com/home-assistant/core/pull/105211


canton7:
Hmph, looks like we can only pass `frozen=True` if the base class is frozen. This probably depends on #480 then.

PR #502, opened 2024-01-12
Fix Load Energy Today Range

mschae:
It was pretty cold here yesterday and I charged my EV therefore easily exceeding 100kWh load and grid consumption.

Hence this fix for the sensors "Grid Consumption Today" and "Load Energy Today".


canton7:
Thanks! I guess we might as well go up to 1000, I'll change.

PR #540, opened 2024-02-15
Try and fix remote control support for H3 and KH

canton7:
Neither has any registers above the active/reactive power control, and the H3 has 32-bit values for active/reactive power.


jaal2001:
If you need to test someone this with a H3-10 let me know.

canton7:
@jaal2001 if you can give the latest beta a test, I'd appreciate it!

jaal2001:
@canton7 Seems to work so far (as one of the previous beta was not able to communicate with my inverter). Only battery can't be tested, as I don't own one.

PR #572, opened 2024-03-19
Update entity_descriptions.py

FozzieUK:
Refactoring introduced error in H1 INPUT load register (should be 11023)


canton7 (changes requested):
Beat me to it, thanks!

canton7 (approved):

PR #574, opened 2024-03-20
Update entity_descriptions.py H3 Battery Charge

FozzieUK:
Refactoring issue, H3 Battery Charge register should be 31036


canton7:
Ack, I gave this refactor two reviews and thought I'd caught everything else.... Thanks for the PRs!

PR #590, opened 2024-04-12
Remove unused PV1 & PV2 power registers from H1_G2 

paultweedy77:
Although specified as such these don’t seem to be 32-bit registers and reading the register pair results in out-of-band values polluting the entities when MPPT is waking up or shutting off.


canton7:
See https://github.com/nathanmarlor/foxess_modbus/discussions/553#discussioncomment-9100039

Issue #599, opened 2024-04-22
Firmware versions reading 0.0 since latest update on invertor H1-6.0


lazenby444:
Version: Master changed to 0.00
11:42:32 - Yesterday
Version: Manager changed to 0.00
11:42:32 - Yesterday
Version: Slave changed to 0.00
11:42:32 - Yesterday

Ndrinta:
Yep, can confirm.

![image](https://github.com/nathanmarlor/foxess_modbus/assets/4704172/f43ad5eb-c328-4a4e-86a9-cd5831e4d6df)

![image](https://github.com/nathanmarlor/foxess_modbus/assets/4704172/1d298dc9-2f4e-4285-a591-d22d8b519949)

![image](https://github.com/nathanmarlor/foxess_modbus/assets/4704172/3171bbd0-dd8a-4d66-a5ac-89bbcb8d48b4)

PR #600, opened 2024-04-23
Fixes firmware version for H1 (G1)

FozzieUK:
Incorrect input registers for Master, Manager, Slave firmware versions on H1 (G1), holding register values are correct.

See https://github.com/nathanmarlor/foxess_modbus/issues/599


canton7:
Thanks!

PR #671, opened 2024-09-03
Fix for Force Charge not working when using H1 over LAN

FozzieUK:
It looks like a bug has crept in on the later releases when trying to set force charge for H1 using LAN.

H1 LAN doesn't support work mode hence it is not possible to set a fall back mode and the 'assert if work_mode_map is not None' correctly raises an error for H1 LAN.

### DIFF
--- a/custom_components/foxess_modbus/remote_control_manager.py
+++ b/custom_components/foxess_modbus/remote_control_manager.py
@@ -272,10 +272,10 @@ class RemoteControlManager(EntityRemoteControlManager, ModbusControllerEntity):
     async def _enable_remote_control(self, fallback_work_mode: WorkMode) -> None:
         # We set a fallback work mode so that the inverter still does "roughly" the right thing if we disconnect
         # (This might not be available, e.g. on H1 LAN)
-        if fallback_work_mode is not None and self._addresses.work_mode_map is not None:
+        if fallback_work_mode is not None and self._addresses.work_mode is not None and self._addresses.work_mode_map is not None:
             fallback_work_mode_value = self._addresses.work_mode_map[fallback_work_mode]
             current_work_mode = self._read(self._addresses.work_mode, signed=False)
-            if current_work_mode != fallback_work_mode_value and self._addresses.work_mode is not None:
+            if current_work_mode != fallback_work_mode_value:
                 await self._controller.write_register(self._addresses.work_mode, fallback_work_mode_value)
 
         if not self._remote_control_enabled:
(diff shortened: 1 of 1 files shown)


FozzieUK:
See https://github.com/nathanmarlor/foxess_modbus/discussions/670

custom_components/foxess_modbus/remote_control_manager.py
```diff
@@ -272,12 +272,16 @@ async def _update_discharge(self) -> None:
     async def _enable_remote_control(self, fallback_work_mode: WorkMode) -> None:
         # We set a fallback work mode so that the inverter still does "roughly" the right thing if we disconnect
         # (This might not be available, e.g. on H1 LAN)
-        assert self._addresses.work_mode_map is not None
-        fallback_work_mode_value = self._addresses.work_mode_map[fallback_work_mode]
+        if (
+            fallback_work_mode is not None
+            and self._addresses.work_mode_map is not None
+        ):
+            assert self._addresses.work_mode_map is not None
```
canton7, line 276: Assertion is superfluous

custom_components/foxess_modbus/remote_control_manager.py
```diff
@@ -272,12 +272,16 @@ async def _update_discharge(self) -> None:
...
+            fallback_work_mode_value = self._addresses.work_mode_map[fallback_work_mode]
 
-        current_work_mode = self._read(self._addresses.work_mode, signed=False)
-        if current_work_mode != fallback_work_mode_value and self._addresses.work_mode is not None:
-            await self._controller.write_register(self._addresses.work_mode, fallback_work_mode_value)
+            current_work_mode = self._read(self._addresses.work_mode, signed=False)
+            if current_work_mode != fallback_work_mode_value and self._addresses.work_mode is not None:
```
canton7, line 280: Probably worth brining the second part of this up to the outer `if`?

canton7:
Thank you!

custom_components/foxess_modbus/remote_control_manager.py
FozzieUK: ahh mypy throws an error about mistmatched arguments if you don't include this check

canton7: I was thinking:

```py
       if fallback_work_mode is not None and self._addresses.work_mode is not None and self._addresses.work_mode_map is not None:
            fallback_work_mode_value = self._addresses.work_mode_map[fallback_work_mode]
            current_work_mode = self._read(self._addresses.work_mode, signed=False)
            if current_work_mode != fallback_work_mode_value:
```

FozzieUK:
@canton7 sorry that was a bit of a chore for such a small change, linting finally passes….. (I’ve tested to confirm it works as expected) - over to you ☺️

canton7:
Thanks! Apologies for being so slow, I'm back now :)

canton7 (approved):

PR #694, opened 2024-10-07
Fix for InvBatPower2 register endian change

FozzieUK:
the latest H3 Pro firmware Master 1.53, Manager 1.22 has flipped the endian for this register; which now matches the Fox modbus definition (V1.05.00.00) see https://github.com/nathanmarlor/foxess_modbus/discussions/685#discussioncomment-10811413


canton7:
Thanks!

PR #696, opened 2024-10-17
Add entities for pv5 and pv6 for H3-Pro inverter

dagroe:
The H3-Pro provides `pv5` and `pv6` in addition to the available `pv1-4`. This PR adds the entities for `pv5` and `pv6` and amends the list of voltages for the H3-Pro.

It also correct the value of `pv4_power` which should be `39285`, based on `pv1_power`: `39279 + 2 * (4 - 1)`

I only added it for `Inv.H3_PRO`, not sure if `Inv.KH_SET` is the same.


canton7:
Looks good, thanks! 

I don't think the KH's have that many strings.

PR #703, opened 2024-11-03
Add SOH for some inverters, implements #687

MartB:
As discussed in #687, this should roughly implement what we know about the SOH registers.

Tested and working on my H3
Master: 2.01; 2.04
Manager: 1.80; 1.81

To Test:
- [x] H3
- [ ] H3-Pro (likely to work)
- [x] H1_G1 :no_entry_sign: (Not supported, confirmed by https://github.com/nathanmarlor/foxess_modbus/discussions/686#discussioncomment-11229454)
- [x] H1_G2 (Confirmed by https://github.com/nathanmarlor/foxess_modbus/pull/703#issuecomment-2453604400)
- [x] KH Series (Confirmed by https://github.com/nathanmarlor/foxess_modbus/pull/703#issuecomment-2588311314)

PR #704, opened 2024-11-03
H3: add bms cell temps and kwh remaining

MartB:
cell_mv_[low|high] are provided in V * 10 by the inverter which makes them unusable without refactoring.

If someone has an idea how to approach a refactor and feels like this is worth it desipite of the low value resolution of 100mv, feel free to guide me into the right direction.

Tested on H3 with
Master: 2.01
Manager: 1.80

### DIFF
--- a/custom_components/foxess_modbus/entities/entity_descriptions.py
+++ b/custom_components/foxess_modbus/entities/entity_descriptions.py
@@ -2131,22 +2131,27 @@ def _bms_entities() -> Iterable[EntityFactory]:
         bms_cell_temp_high=[
             ModbusAddressesSpec(input=[11043], models=Inv.H1_G1 | Inv.KH_PRE119),
             ModbusAddressesSpec(holding=[37617], models=Inv.H1_G2_144 | Inv.KH_PRE133 | Inv.KH_133),
+            ModbusAddressesSpec(holding=[31102], models=Inv.H3_SET),
         ],
         bms_cell_temp_low=[
             ModbusAddressesSpec(input=[11044], models=Inv.H1_G1 | Inv.KH_PRE119),
             ModbusAddressesSpec(holding=[37618], models=Inv.H1_G2_144 | Inv.KH_PRE133 | Inv.KH_133),
+            ModbusAddressesSpec(holding=[31103], models=Inv.H3_SET),
         ],
         bms_cell_mv_high=[
             ModbusAddressesSpec(input=[11045], models=Inv.H1_G1 | Inv.KH_PRE119),
             ModbusAddressesSpec(holding=[37619], models=Inv.H1_G2_144 | Inv.KH_PRE133 | Inv.KH_133),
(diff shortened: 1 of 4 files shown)


MartB:
Will work on this again after #703 goes in.

canton7:
> Will work on this again after #703 goes in.

#703 is now in!

MartB:
@canton7 thanks! Any idea on how to make this scaling / conversion happen without a lot of changes to the current code?

FozzieUK:
Just to add my comments here as at 100mV resolution these values are virtually useless, the Fox protocol specs have always shown this as 0.1V resolution but the H1 (Gen1) inverters provide this in mV ~ which is what the BMS<->Battery CAN bus protocol transmits as part of the pack status message.

At 0.1V resolution the SoC difference could be > 50% and the imbalance calculation would be utterly useless (a maximum of 40mV swing between low and high cell voltage is significant, at 100mV the pack is useless).

My working theory is Fox got this wrong in the original protocol and the 'error' was carried over to the latest protocol spec. That has been somewhat proven in the latest H1 G2 firmware that uses the latest spec *but* in the last firmware release in a break with the spec Fox have changed the BMS cell voltages to report mV again - I am hoping this change gets rolled out to their other inverters in the near future.

In terms of scaling, the entity can be modified to select scale, it will need a seperate yield statement for each inverter model and possibly  the sub firmware versions (as is the case with the G2 inverter) - there are several examples of this already (pv_current, grid_ct, inv_current etc..) and there will be more in future - the H3 Pro latest release has uniquely changed the scale on all of the total registers.

My only real question on the BMS cell voltages is what benefits would there be if the scale was corrected so that it reported at 0.1V resolution - I can't see many ? - a simple translation can be handled with a template in home assistant.

AND I still remain hopeful that Fox will correct the resolution in other inverters, but that only remains a hope.

MartB:
@FozzieUK Thanks for the valuable insights!

I will leave the voltages out but keep the comment with the registers.
@williamjeccles do you happen to have any insights into why the cell voltages are represented in modbus like this?

Will prepare the remaining kwh (works together quite well with SOH) and cell temps today and make this available to merge.

MartB:
@canton7 @FozzieUK should be g2g.

canton7:
Thanks!

canton7:
v1.13.1b3 it is. Tbh it *should* be stable, and the more people who try it the better, so I haven't labelled it as invite only.

FozzieUK:
@canton7 ok thanks, i've tried it on 2025.1.4 and 2025.2.0b4 and stable for me - looking very good 👍

canton7:
Thank you!

canton7:
Ack I replied in the wrong tab sorry -- should have been in #749

canton7:
I've had to revert this in #759, see #756

canton7:
The discussion on re-adding these is in #760

PR #730, opened 2025-01-16
AC1 G2 inverter profile added

FozzieUK:
Add AC1 G2 has same register mapping as H1 G2

see https://github.com/nathanmarlor/foxess_modbus/discussions/715


canton7:
Thanks!

PR #731, opened 2025-01-16
Add Peak Shaving work mode for H3 Inverters & H1 (Gen 2)

FozzieUK:
Adds the Peak Shaving work mode in the select  for H3 set Inverters & H1 (Gen 2) 

See https://github.com/nathanmarlor/foxess_modbus/discussions/699

### DIFF
--- a/custom_components/foxess_modbus/entities/entity_descriptions.py
+++ b/custom_components/foxess_modbus/entities/entity_descriptions.py
@@ -2104,7 +2104,7 @@ def _configuration_entities() -> Iterable[EntityFactory]:
     yield ModbusWorkModeSelectDescription(
         key="work_mode",
         address=[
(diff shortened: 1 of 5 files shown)


FozzieUK:
The pytest is failing on work mode even though the snapshots appear correct ?

FozzieUK:
Always fails on Inv.H3 pytest --snapshot-update deletes the H3 file and doesn't recreate with the following error ?

FAILED tests/test_entity_descriptions.py::test_entity_descriptions_for_model[Inv.H3] - RuntimeError: dictionary keys changed during iteration

FozzieUK:
As i've been trying (without success) to get the pytest to succeed I'm going to close this pull request and create a new one with just the change and the revised tests - and then hopefully someone will be able to advise on how to overcome the failing inv.H3 test (all the other inverter model tests are ok?).

PR #737, opened 2025-01-19
Add Peak Shaving work mode for H3 Inverters & H1 (Gen 2)

FozzieUK:
Adds the Peak Shaving work mode in the select for H3 set Inverters & H1 (Gen 2)

See https://github.com/nathanmarlor/foxess_modbus/discussions/699

### DIFF
--- a/custom_components/foxess_modbus/entities/entity_descriptions.py
+++ b/custom_components/foxess_modbus/entities/entity_descriptions.py
@@ -2095,7 +2095,7 @@ def _configuration_entities() -> Iterable[EntityFactory]:
         key="work_mode",
         address=[
             ModbusAddressSpec(input=41000, models=Inv.H1_G1 | Inv.KH_PRE119),
-            ModbusAddressSpec(holding=41000, models=Inv.H1_G1 | Inv.H1_G2 | Inv.KH_119 | Inv.H3_SET),
+            ModbusAddressSpec(holding=41000, models=Inv.H1_G1 | Inv.KH_119),
         ],
         name="Work Mode",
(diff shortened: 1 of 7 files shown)


FozzieUK:
If anyone can assist with this -

This is for a small change to the workmode select for the H1_G2 and H3_SET it has been tested in operation ok but when doing the PR all of the tests are ok with the exception of pytest on the Inv.H3 - when I run a pytest --snapshot-update it fails with this error (as if the test script itself is breaking) but only on the Inv.H3 ?

FAILED tests/test_entity_descriptions.py::test_entity_descriptions_for_model[Inv.H3] - RuntimeError: dictionary keys changed during iteration

FozzieUK:
To avoid the runtime error i've had to make a change in test_entity_descriptions.py, the pytest snapshot now works correctly and the output in the inverter model tests is correct.

FozzieUK:
@canton7 Hi Antony would you please review this change - i've had a load of problems for what was only a minor change to the workmode select for the H3_SET to add 'Peak Shaving', and the pytest snapshots start to fail with a RuntimeError: dictionary keys changed during iteration.

When I looked more closely at the content of the existing tests I found that with work_mode select the test results were random and in the case of the H3_Pro incorrectly showing only 2 of the options - in the case of the Inv.H3 the snapshot crashed with the above error.

I've had to make a change in `test_entity_descriptions.py` changing from `for k, v in d.items():`  to `for k, v in d.copy().items():` and the snapshots now run correctly (and i've verified each manually) but I would be grateful if you could check my logic !

canton7:
Good catch on the test!

Thanks, looks good.

PR #735, opened 2025-01-19
Adds additional BMS entities for H1 G2 and KH inverters on latest firmware

FozzieUK:
Adds additional BMS entities for H1 G2 and KH inverters running with latest firmware

On H1 G2 inverter this PR will add these entities-

bms_cell_temp_high
bms_cell_temp_low
bms_cell_mv_high
bms_cell_mv_low
batvolt
bat_current
inverter_fault_code

Note for the G2 this update is sensitive to firmware version:

On the H1 G2, if you have Manager: 1.37, Master: 1.22 it will not support all the BMS registers - either upgrade to the latest firmware or in the integration configure the settings and enter the manager version of 1.44
also note Manager firmware 1.44 has incorrect scale on bms_cell_mV, low/high (this is a Fox bug, upgrade to Manager 1.48 to fix this)

see https://github.com/nathanmarlor/foxess_modbus/discussions/712

On KH inverter this PR will add these missing BMS entities-

bms_cell_temp_high
bms_cell_temp_low
bms_cell_mv_high
bms_cell_mv_low
bms_kwh_remaining

see https://github.com/nathanmarlor/foxess_modbus/pull/740#issuecomment-2609903770

Issue #661, opened 2024-08-15
K series unable to Force Charge/Discharge


williamjeccles:
My setup is two KH10.5 wired in parallel (master/slave enabled) and connected to full stacks of HV2600 8* batteries per inverter.
The inverter and batteries are all on the very latest firmware using the latest version of HA and the modbus plugin.

If I set a force discharge/charge period via the Fox Cloud web UI the inverters are able to discharge/charge without issue, no fighting or flapping.

However, I can reproduce the issue if I use HA via the plugin to set the inverters into Force Charge workmode or Force Discharge workmode the inverters start flapping and fighting as per the video example.

https://youtu.be/1LQeiYCD9vs - Repro example video

I have a suspicion that the kW limit is causing the issue for systems that are parallel. Almost like its telling the systems to discharge at say 7kW total instead of per inverter.

![image](https://github.com/user-attachments/assets/ef20c9b7-d7ae-4ca4-8592-91c20c036880)

I would like to bypass the UI limit as I believe this might be causing the inverters to flap/fight.
So I'm happy to test and rule out this suspicion if someone could point me which file and line I should comment or adjust to test out my theory.

I believe if I could set the force discharge limit to 21kW then it would allow both inverters to max out their charge/discharge amount instead of fight. It's not a perfect fix but at least it would unblock and validate my current belief.

williamjeccles:
@canton7 we discussed this a while back but would like to reopen the topic and see if we can solve it for all parallel users.

FozzieUK:
That's an interesting one, with parallel inverters do you set one of them to force charge/discharge and then they 'share' the output ? - as this might be a bit more complex to unravel.

Whilst we're waiting on Antony, it's a quick change to hard code the limit and test the theory, the file you are looking to change is 

https://github.com/nathanmarlor/foxess_modbus/blob/main/custom_components/foxess_modbus/inverter_profiles.py


I think a change in line 77 is all that is needed to test with i.e.


Change
```
            capacity = int(float(capacity_str) * 1000)
```

To this
```
            capacity = int(21000)
```

williamjeccles:
> That's an interesting one, with parallel inverters do you set one of them to force charge/discharge and then they 'share' the output ? - as this might be a bit more complex to unravel.
> 
> Whilst we're waiting on Antony, it's a quick change to hard code the limit and test the theory, the file you are looking to change is
> 
[...]

PR #740, opened 2025-01-22
Add KH version 1.33 (thanks to @ocl80)

canton7:
Hi @ocl80,

Thanks for carrying the torch on the KH v1.33 while I got version selection sorted out.

I've done a squash-merge of your fork into this MR (with you as the author), then a couple of fixup commits on top to avoid breaking other inverter types. I hope that's OK. Can you just check that it looks sensible please?

Thanks!


ocl80:
Thanks @canton7 - for spotting I had failed to revert my earlier (crude) fix which included changing the pv current scale factor.
I had not understood how the post_process lambda function to switch the grid_ct polarity was applying only to the KH... now I see it wasn't and there was simply a negative scale factor that I had missed, and which is not needed now for KH133. 
All good for me thank you!

WyndStryke:
This is awesome, thank you both :-)

Any chance of getting the BMS cell low temp into KH133, if it isn't already?  It seems to work when I read register 37618.  I think that'd allow predbat to use the temperature charge curve functionality on the KH (although right now it still seems to be a bit buggy on the predbat side).

canton7:
@ocl80 Thank you, and thanks for polishing the KH support! Yep, the snapshot tests came into their own there.

@WyndStryke That's #735 right? That needs a little update to support the version-selector stuff I got in today. I'll take a look on Sunday unless @FozzieUK gets there first.

WyndStryke:
> ... That's #735 right? That needs a little update to support the version-selector stuff I got in today. I'll take a look on Sunday unless @FozzieUK gets there first.

Ah looks like the registers are the same, so yes (I had assumed they'd be different)

williamjeccles:
> --- Edit: 49203 (work mode) does not update when force charging is active, it stays at 1 (I was expecting 6 for force-charge). Interestingly the Fox app also shows self-use despite also showing the battery being charged from grid (via the mode scheduler). So I'm not sure what's going on there, maybe a firmware bug reporting the wrong workmode, or perhaps it's actually a write-only field? Is there another way to retrieve the workmode?

I believe I came across the force charge issue, the reason for it and a solution here https://github.com/nathanmarlor/foxess_modbus/issues/661

WyndStryke:
Just ploughing through that thread, thank you :-)  It's going to take a while because it's a bit beyond my current knowledge.

I had seen it a while ago, but I had assumed it just related to parallel setups and ignored it.

Regarding

> As I recall there may be a problem with reading the Remote Take effect register as the KH did not expose registers above 44003 - it would be useful to know if this is still the case and needs to be re-checked on a KH

I seem to be able to read registers such as 49222, 49079 OK (KH > 1.33)

Regarding **reading** the current work mode via 49203, I guess that firstly the mode scheduler doesn't use that mechanism, and instead uses remote control?  So presumably I'd need to read the 44xxx registers instead to see if the inverter had remote control turned on, and what it had been asked to do?

Right now, with self-use active, they read:

* 44000 1
* 44001 900
* 44002 0

Which does actually correspond with the final force charge segment in my schedule (6:45-6:59, 10% min soc, 80% max soc) which is basically there just to hold the charge steady in case of time drift.  I don't have any force discharges configured because I don't have an export MPAN / MCS yet (still in the paperwork stage).

FozzieUK:
> I'll go through and test each of them to confirm. This is on an EC4300-H4 (usable capacity 15kWh) and a KH7, manager 1.37, master 1.41, slave 1.01.
> 
> * BMS1 MaxTemp 37617 147
> * BMS1 MinTemp 37618 105  (Fox app 10.5)
> * BMS1 MaxCellVolts 37619 32   (not sure how to check this, SoC is 63%, 37609 BMS1 voltage is 2352)
> * BMS1 MinCellVolts 37620 32
> * BMS1 SOH 37624 100  (fox app 100%)
> * BMS1 RemainEnergy 37632 1658   (nominal capacity 16590Wh)
> * **BMS1 FCC Capacity 37633 0**
> * **BMS1 DesignEnergy 37635 0**
> 

Ok thanks, those bms temp and cell registers are all working as per the latest modbus spec - I'll look at modifying my PR for the G2 and add KH BMS support.

> 
> Could 'FCC Capacity' be related to 'Number of **f**ull equivalent **c**harge-discharge **c**ycles' in the fox app?

These don't seem to be populated yet in any tests we've done, it appears to be future development scope.

> 
> A few other random observations while I was poking around with registers:
> 
> * 49203 work mode - returns 1 which is self-use (& what I have set during the day).  What I've noticed is that this doesn't change in the integration despite swapping to force charge overnight on the inverter via the mode scheduler.  I'll have to re-query this overnight to see if it picks up the force charge.  Looking at the source I'm not sure where it is coming from at the moment for the KH133 profile.
> --- Edit: 49203 (work mode) does not update when force charging is active, it stays at 1 (I was expecting 6 for force-charge). Interestingly the Fox app also shows self-use despite also showing the battery being charged from grid (via the mode scheduler). So I'm not sure what's going on there, maybe a firmware bug reporting the wrong workmode, or perhaps it's actually a write-only field? Is there another way to retrieve the workmode?

You won't see the inverters work mode register change when you force charge (or discharge) - these modes are created internally by the integration using the remote power registers (the inverter work mode is unchanged and will be in whatever work mode it was left in)

If you want to know the integration work mode you will have to test the state of `select.work_mode` which will report the text Self Use, Feed-in First, Backup, Force Discharge, Force Charge etc..

canton7:
This PR is going to be closed shortly. Please don't use it for general conversation: that will be lost. I suggest an issue / discussion instead!

PR #746, opened 2025-01-26
Add BMS registers to new KH firmware

canton7:


canton7:
See https://github.com/nathanmarlor/foxess_modbus/discussions/745#discussioncomment-11962609

Issue #748, opened 2025-01-30
Do not upgrade to HA 2025.2. We'll get a fix out when we can.


stevetrease:
Starting Home Assistant 2025.2 RC and foxess_modbus v1.13.0 results in an "invalid config" notification in Home Assistant.

Let me know if I can provide more information or help with testing a resolution....

## Version of the custom_component
- foxess_modbus v1.13.0 
- Home Assistant Core 2025.2.0b1

## Describe the bug
Starting Home Assistant 2024.2 RC and foxess_modbus v1.13.0 results in an "invalid config" notification in Home Assistant. 
```
Invalid config
The following integrations and platforms could not be set up:
foxess_modbus
Please check your config and [logs](https://hatest.tail5e5d.ts.net/config/logs).
```

## Debug log
Logger: homeassistant.setup
Source: setup.py:334
First occurred: 14:27:14 (1 occurrences)
Last logged: 14:27:14

Setup failed for custom integration 'foxess_modbus': Unable to import component: No module named 'pymodbus.pdu.register_read_message'
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/setup.py", line 334, in _async_setup_component
    component = await integration.async_get_component()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/loader.py", line 1035, in async_get_component
    self._component_future.result()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/src/homeassistant/homeassistant/loader.py", line 1015, in async_get_component
    comp = await self.hass.async_add_import_executor_job(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        self._get_component, True
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/src/homeassistant/homeassistant/loader.py", line 1075, in _get_component
    ComponentProtocol, importlib.import_module(self.pkg_path)
                       ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/util/loop.py", line 201, in protected_loop_func
    return func(*args, **kwargs)
[...]

PR #749, opened 2025-01-31
Vendor pymodbus, rather than relying on the version which HA installs

canton7:
This should protect us from HA breaking things in the future!

Fixes: #748


FozzieUK:
I've been (over) thinking about pymodbus versions and which is closest to what we had prior to the change in core 2025.1.x. 

There have been a few issues reported since the change to 3.7.4, a lot because of install and other modbus dependencies but a few systems worked ok on core 2024.12.x and not on 2025.1.x (i'd assumed 3.7.4 has specific issues).

Before the change to handle core 2025.1.x the manifest specified >= 3.1.3 and requirements.txt ==3.5.4 so i'm guessing? the dependencies in requirements would have won out but it stopped working because of the api changes in 3.7.4 so it must have been using that.

The actual version of pymodbus shipped prior to 2025.1.x was 3.6.9 and that seemed to be as good as any - so I wondering if it would be better to use 3.6.9 as the vendor specific version ?

canton7:
Good pont. Home assistant used the following versions:

* Jul 2024: 3.6.9
* Apr 2024: 3.6.8
* Apr 2024: 3.6.7
* Mar 2024: 3.6.6
* Mar 2024: 3.6.5
* Feb 2024: 3.6.4
* Jan 2024: 3.6.3
* Oct 2023: 3.5.4
* Sep 2023: 3.5.2
* Sep 2023: 3.5.1
* Aug 2023: 3.5.0
* Aug 2023: 3.4.1
* Jun 2023: 3.3.1

So we haven't actually stayed on any particular version of pymodbus for any length of time at all.

Looking back through the logs, 3.5.4 was the last time I checked to see whether our patches on top of pymodbus were correct. But as you say, 3.6.9 seemed to be working fine. Either way, 3.1.3 is probably a bit old :P

canton7:
3.6.9 it is!

FozzieUK:
Thanks, ready to test 👍

canton7:
@FozzieUK Do you want any itme to test before I push out a release?

FozzieUK:
It'll be easier if you can push it out as a pre-release and then I can try it on a few different systems - i've been lucky so far and not had any problems using the waveshare so will be good to expand it out and find a few candidates that have had issues and get them to test first.

Could you label it 'Don't use this unless invited' or something

PR #750, opened 2025-01-31
Fix detection of AC1-G2

canton7:
The original patch had duplicate InverterModel keys in INVERTER_PROFILES, meaning that one or the other would be kept.

See: #715


canton7:
@FozzieUK This one's for you!

FozzieUK:
Doh............

Thank you 👍

Issue #756, opened 2025-02-01
Kuara H3, and some KHs and H3s doesn't support new registers in 1.13.1


Badlee2020:
With version 1.13.2b1 I get the following error:

Error reading registers. Type: RegisterType.HOLDING; start: 31000; count: 91; slave: 196 from tcp://192.168.2.235:502: Exception Response(131, 3, IllegalAddress)

With version 1.13.0 every was ok

Manager 1.62
Master 1.74
Slave 1.03

I have a  KOPP-KUARA 12.0-3H

canton7:
Right, I see. It looks like #756 added support for battery SoH, but this isn't supported by the Kuara. I'll fix.

canton7:
Can you just check quickly whether you can read holding registers 31102, 31103, 31123? Developer tools -> Actions -> FoxESS - Modbus: Read Registers.

canton7:
And do you know whether the Kuara supports Peak Shaving work mode?

Badlee2020:
31102, 31103, 31123 these registers produce errors. Peek shaving is not a work mode I have seen. More than willing to test it though:) the documentation with Kopp is quite 😔

jbarrett-fmad:
FYI I'm getting the same error on KH10.5 Manager version 1.21. Have requested firmware update from Fox to see if that resolves it.

hooktheboss:
H3-6.0-E - AUX no connection after Update to v1.13.1

Error connecting to inverter (friendly name: "FoxESSInverter"). Error reading registers. Type: RegisterType.HOLDING; start: 31000; count: 91; slave: 247 from tcp://192.168.178.17:512: Exception Response(131, 3, IllegalAddress)

Manager 1.56
Master 1.67
Slave 1.03

Reverted to v1.13.0 - working OK.

AndaleR:
Same here.

Worst case: Will all data be disappeared? Or is there a solution and it just needs to be updated and all the energy-data will be back in HomeAssistant.

canton7:
Right, I've releaed v1.13.2 which reverts those changes. I'll get them inot the next beta, and we can work out what firmware versions actually support them.

iancg:
> Right, I've released v1.13.2 which reverts those changes. I'll get them back into the next beta, and we can work out what firmware versions actually support them.

Would another approach be to try reading those registers as a test read after startup, and if success then add them in to the map, otherwise log and skip them? 

[...]

PR #757, opened 2025-02-01
Remove Battery SoH from the Kuara H3

canton7:
Fixes: #756


canton7:
FYI @MartB. Sorry, I should have spotted that you hadn't tested this.

Issue #772, opened 2025-02-03
KH firmware pre_133 does not support BMS registers


FozzieUK:
## Configuration

KH10 running firmware Manager 1.22, Master 1.39, Slave 1.01

Taken from the Facebook group where a KH user is struggling to get system restarted, I tried getting him to set the config version to 1.33 but the BMS registers do not exist on this firmware.

https://www.facebook.com/groups/foxessownersgroup/permalink/2280867245628778

## Describe the bug

KH firmware PRE133 does not support BMS registers cell_temp_high, cell_temp_low, cell, cell_mV_high, cell_mV_low, kwh_remaining - the addresses 37617, 37618, 37619, 37620, 37632 and need to be removed for Inv.KH_PRE133

## Debug log

```text

Logs taken from a KH10 running firmware Manager 1.22, Master 1.39, Slave 1.01

Connection Error
Error ⸱ Reported by FoxESS - Modbus
Error connecting to inverter (friendly name: ""). Error reading registers. Type: RegisterType.HOLDING; start: 37617; count: 16; slave: 247 from tcp://192.168.1.254:502: Exception Response(131, 3, IllegalAddress)

```

PR #773, opened 2025-02-03
KH firmware pre_133 does not support BMS registers

FozzieUK:
See https://github.com/nathanmarlor/foxess_modbus/issues/772


canton7:
Huh, it would be worth finding what PR added this - it was obviously deliberate. I wonder whether we need to split PRE113 up further

FozzieUK:
Stupidly I added it as an amendment to my H1 G2 BMS registers, it came from the KH133 thread / where users said it supported it - I won’t take anyones word unless I double check next time…

canton7:
Cool, I'll take your word for it, thanks.

PR #779, opened 2025-02-14
Splits AIO_H3 into 2 versions than can be selected by manager version

FozzieUK:
Splits AIO_H3 into 2 manager versions than can be selected by the manager version.

AIO_H3_PRE101 is for older AIO_H3's that are unable to support the following entities:
`workmode, max_charge_current, max_discharge_current, min_soc, min_soc_on_grid, max_soc,`

In Settings -> Devices & Services -> FoxESS - Modbus -> Configure -> Version settings (select 1.1)

AIO_H3_101 has the same entities as before i.e. H3_SET without remote control.

(https://github.com/nathanmarlor/foxess_modbus/discussions/776) refers


canton7:
Looks good, thanks!

PR #830, opened 2025-04-21
Fix config flow strings

andrew-codechimp:
Fix some missing config flow strings, one had an incorrect key as well


canton7:
Thanks!

PR #840, opened 2025-05-09
H3 Pro - scale changed for daily and total energy entities

### LINKED ISSUE #724: H3 Pro Latest Inverter firmware changes totals x 10
397998058-94c4b46e-fbd5-41ef-bc63-4f1a83c26ae6

H3 pro 30 KW updated firmware to 1.25 and 1.56 and battery charge and discharge total has jumped by a decimal place 4800 to 48000. Feed in total, load energy and yield have also jumped. by the same
[Uploading Fox.log.txt…]()

Version of the custom_component
1.12.2

andyjenkinson:
Fixes #724

Later versions of H3 Pro firmware use a scale of 0.01 for the daily and running total energy values (e.g. `solar_energy_total`) instead of 0.1.

This appears to match the behaviour of H3 Smart model.

I have assumed the changes became effective as of firmware dev release 1.22, this seems likely and a model profile already exists for post-1.22 versions so re-using that simplifies the fix. We know from #724 that it appeared in 1.25 at the latest, and remains in versions 1.33 and 1.34.

It works for me but I don't want to break it for anyone on a previous firmware version, so I would like someone on 1.22 to confirm if the scale is off on theirs too. If  not then it must have been changed _after_ 1.22 - in that case I'll change it so it applies  from 1.25...

### DIFF
--- a/custom_components/foxess_modbus/entities/entity_descriptions.py
+++ b/custom_components/foxess_modbus/entities/entity_descriptions.py
@@ -1563,13 +1563,13 @@ def _inverter_entities() -> Iterable[EntityFactory]:
             ModbusAddressesSpec(
                 holding=[32001, 32000], models=Inv.H1_G1 | Inv.H1_G2_SET | Inv.H3_SET | Inv.KH_PRE133 | Inv.KH_133
             ),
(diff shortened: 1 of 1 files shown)


canton7:
Thanks!

I have some notes saying that the format of the master/slave version changed (to hex). Is this true?

PR #846, opened 2025-05-16
Feature/reactive power

andyjenkinson:
Added measurements for:
- Grid CT reactive power
- Grid CT apparent power
- Grid CT power factor

Sensors for combined and for each phase. I'm unsure if positive and negative are the same way around as the existing (active) power measurements.

I am unsure if these are supported by earlier versions of the H3 Pro, may need a tweak to the model profile if these fail for earlier versions.

I also assumed that they are supported by H3 Smart since they seem to be following the same modbus spec, but I can't test.


canton7:
After some bad experiences assuming that older inverters supported things that they didn't, I'm going to be conservative and just enabled this for inverters we've tested it on.

PR #866, opened 2025-06-24
H3_180: use proper cell mv registers

MartB:
Somehow I overlooked that these registers exist. Found them while scanning for new stuff in 1.87. They are in mV! I also found the export limit but that's probably known already.

Version is: 1.87, can someone verify they exist on 1.80?


canton7:
Thanks! Are you saying the old register addresses are incorrect?

MartB:
https://github.com/nathanmarlor/foxess_modbus/commit/58c26d26bb5cec0b58465cfbe00f81e8927b6bbb

Was here but got removed when the code was reverted partially and redone. So we are g2g.

PR #916, opened 2025-08-31
Adding support for P1 Inverters

migueldmrodrigues:
Was trying to integrate my P1 Inverter and found that there was a discussion already where a user confirmed it was the same as H1-G2 so I somewhat copied this PR: https://github.com/nathanmarlor/foxess_modbus/pull/730/files, tested on my instance and worked.

Resolves: https://github.com/nathanmarlor/foxess_modbus/discussions/914

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Supported models:
 - FoxESS H1 (including AC1, AIO-H1 and G2)
 - FoxESS H3 (including AC3 and AOI-H3)
 - FoxESS H3 PRO
+- FoxESS P1
 - FoxESS KH
 - Kuara H3
 - Sonnenkraft SK-HWR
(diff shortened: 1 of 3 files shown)


nathanmarlor (changes requested):
Thanks for adding P1 support. A few minor changes needed before this can be merged:

1. **Indentation fix.** The `InverterModelProfile(` line has 5 spaces of indentation instead of 4 — please fix to match the rest of the file.

2. **Missing test snapshots.** Run the test suite with `--snapshot-update` and include the generated P1 snapshot files (P1-AUX-latest.json and P1-AUX-v1.44.json).

3. **Regex breadth.** The pattern `^P1-([\d\.]+)-E` requires a `-E` suffix. Are there P1 variants without `-E`? If so, consider broadening the regex.

Once those are addressed this is good to go.

plutec:
Hi @migueldmrodrigues, could you make the required changes to be approved? Otherwise I can make a fork and fix again, but I believe it is better to don't duplicate

PR #934, opened 2025-10-08
Add Export Limit control for KH inverters

mystardious:
- Add export_limit property to EntityRemoteControlManager interface
- Implement export_limit setter in RemoteControlManager with async register write
- Add ModbusRemoteControlNumberDescription for export limit in remote control config
- Restrict export limit to KH models only (KH_PRE119, KH_PRE133, KH_133)
- Use register 41012 for KH inverters
- Support 0-10kW range with 0.001kW steps
- Integrate with existing remote control system

### DIFF
--- a/custom_components/foxess_modbus/common/entity_controller.py
+++ b/custom_components/foxess_modbus/common/entity_controller.py
@@ -80,6 +80,17 @@ class EntityRemoteControlManager(ABC):
     def max_soc(self, value: int | None) -> None:
         """Set a value to override the max_soc register, if any"""
 
+    @property
+    @abstractmethod
+    def export_limit(self) -> int | None:
+        """Get the current export limit"""
+
+    @export_limit.setter
+    @abstractmethod
+    def export_limit(self, value: int | None) -> None:
+        """Set the export limit"""
+
+
 
 class EntityController(ABC):
     """Interface given to entities to access the ModbusController"""
--- a/custom_components/foxess_modbus/entities/modbus_remote_control_config.py
+++ b/custom_components/foxess_modbus/entities/modbus_remote_control_config.py
@@ -179,7 +179,28 @@ class ModbusRemoteControlFactory:
             value_setter=_set_max_soc,
         )
 
+        def _set_export_limit(manager: EntityRemoteControlManager, value: int) -> None:
+            manager.export_limit = value
+
+        # Export limit only for KH models
+        kh_models = [x.get_all_models() for x in self.address_specs if x.models & Inv.KH_SET]
+        export_limit = ModbusRemoteControlNumberDescription(
+            key="export_limit",
+            name="Export Limit",
+            models=kh_models if kh_models else [],  # Only KH models
+            native_max_value_callback=lambda x: x.inverter_capacity,  # Use inverter capacity
+            mode=NumberMode.BOX,
+            device_class=NumberDeviceClass.POWER,
+            native_min_value=0.0,
+            # Max value is read from the inverter
+            native_step=0.001,
+            native_unit_of_measurement="kW",
+            scale=0.001,
+            value_setter=_set_export_limit,
+        )
+
         self.entity_descriptions: list[EntityFactory] = [
+            export_limit,
             charge_power,
             discharge_power,
             remote_control_select,
--- a/custom_components/foxess_modbus/remote_control_manager.py
+++ b/custom_components/foxess_modbus/remote_control_manager.py
@@ -28,6 +28,7 @@ class RemoteControlManager(EntityRemoteControlManager, ModbusControllerEntity):
         self._discharge_power: int | None = None
         self._charge_power: int | None = None
         self._max_soc_override: int | None = None
+        self._export_limit: int | None = None
 
         modbus_addresses = [
             *self._addresses.battery_soc,
@@ -73,6 +74,21 @@ class RemoteControlManager(EntityRemoteControlManager, ModbusControllerEntity):
     @max_soc.setter
     def max_soc(self, value: int | None) -> None:
         self._max_soc_override = value
+
+    @property
+    def export_limit(self) -> int | None:
+        return self._export_limit
+
+    @export_limit.setter
+    def export_limit(self, value: int | None) -> None:
+        self._export_limit = value
+        # Export limit writes to register 41012 (H1, H3, KH series) or 46616 (other models)
+        if value is not None:
+            # Schedule the async write operation using Home Assistant's event loop
+            self._controller.hass.async_create_task(
+                self._controller.write_register(41012, value)
+            )
+
 
     async def _update(self) -> None:
         if not self._controller.is_connected:


mystardious:
Adds another box to allow you to specify the export limit.

Screenshot 2025-10-08 at 12 52 43 pm

MartB:
41012 with u32 also works for H3 after 180 (as far as I can confirm) 

Been using it for a while.

jeffborg:
How hard would it be to add import limit to this as well? Would be useful if you have a 2nd high power device eg ev charger who also has a grid limit so set which device inverter or ev charger gets priority for grid power.

nathanmarlor (changes requested):
This is a great feature and clearly wanted by the community — thanks for the work. However, the implementation has some architectural issues that need addressing:

1. **Fire-and-forget register write.** The `export_limit` setter calls `async_create_task(write_register(41012, value))` directly. Every other setter in `RemoteControlManager` just stores a value and lets the `_update()` cycle handle writes. This bypasses error handling, debouncing, and the remote control lifecycle (e.g., writes still go through even if remote control is disabled).

2. **Hardcoded register address.** The address `41012` is hardcoded in the setter instead of coming from `ModbusRemoteControlAddressConfig`, which is the pattern used for all other register addresses in the remote control system.

3. **Architectural fit.** Export limit is a persistent inverter configuration setting, not a remote control action. It might be better implemented as a standalone writable number entity backed by a holding register, rather than being part of the remote control manager.

4. **Breaking interface change.** Adding an abstract `export_limit` property to `EntityRemoteControlManager` means any other implementors of that interface would break.

5. **Missing test snapshots.**

6. **Scope.** @MartB noted register 41012 also works for H3 after firmware 1.80 — worth considering whether this should be KH-only.

Firespray-31:
Is this functionality not already added in this commit (for KH_133 only at present) that was recently just merged in?

https://github.com/nathanmarlor/foxess_modbus/commit/a2d670ed7de764ef71134a5625006e8d46fd3f0a

MarcFoxEss:
Register 41012 for KH is not working with latest FW - Manager 1.61 Master 1.67

nickw444:
> Is this functionality not already added in this commit (for KH_133 only at present) that was recently just merged in?

Seems so, I switched to `master` and I now have control over export limit

image

nathanmarlor:
Thanks for this — export limit for KH is wanted. A few changes needed before it can merge:

1. **Route the write through the normal lifecycle.** The setter currently does `async_create_task(write_register(41012, value))`, which bypasses the `_update()` flow, error handling, debouncing and the remote-control enable state that every other setter uses. Export limit is really a persistent configuration value, so the cleanest fit is a **standalone writable `number` entity backed by the holding register** (like the H3-Smart/KH export_power_limit entities at 46617/46616) rather than going through `RemoteControlManager`.
2. **Don't hardcode 41012.** Source the register from config rather than inline.
3. **Drop the breaking abstract change.** Adding an abstract `export_limit` to `EntityRemoteControlManager` breaks the other implementors.
4. **Scope:** MartB notes 41012 (u32) also works on H3 ≥ 1.80, so this needn't be KH-only — worth covering both.
5. **Add the test snapshots** (`pytest --snapshot-update`) and clear the stray double blank lines (they'll fail ruff).

Note there's now already an `export_power_limit` number entity for KH_133/H3_SMART (#1076, #1098) — worth checking whether your use case is covered by that, or whether this is a distinct register. Happy to help reshape it as a standalone number entity if that direction works for you.

momomosun:
I can confirm KH has a working export limit over Modbus — but **41012 is not it on my firmware**, which may be why this PR has been hard to land.

Unit: **KH10**, Manager 1.64 / Master 1.69 / Slave 1.03, AUX RS485 → EW41, `rtu_over_tcp`, slave 247.

## TL;DR

Export limit on this KH10 is a **u32 at holding registers 46616 (high) / 46617 (low)**, in **watts**, mirrored at **41022 / 41023**. Writing either pair changes what the FoxESS cloud reports for the `ExportLimit` device setting, and it demonstrably limits real grid feed-in.

## 41012 does nothing here

It reads `100` and only accepts 0–100 — writes of 1000 / 10000 / 10500 / 11000 are all rejected, value stays 100. And it has no effect on feed-in:

| 41012 | load | feed-in | battery discharge |
|---|---|---|---|
| 100 | 3.01 kW | 3.77 kW | 7.12 kW |
| 30 | 3.03 kW | 3.76 kW | 7.12 kW |
| 10 | 1.76 kW | 5.03 kW | 7.12 kW |

Feed-in is fully explained by `discharge − load` in all three rows. Note your own dump shows `41012 = 10000` while mine reads `100`, so the semantics genuinely differ across firmware.

## How I found the real one

Change the setting from the cloud, then diff the registers:

1. Read **every address individually** across 40000-40060, 41000-41060, 42000-42060, 44000-44060, 46600-46660. Per-address reads are essential — 41019-43999 is inside `KH_REGISTERS.invalid_register_ranges`, so those registers never show up in a bulk read. My earlier bulk scans stopped at 41018 and missed 41023 by five addresses.
2. Three registers read exactly my cloud `ExportLimit` value of 15000: **41021, 41023, 46617**.
3. Changed cloud `ExportLimit` 15000 → 12345 and re-scanned: **41023 and 46617 followed, 41021 did not.**
4. Reverse check — wrote `41023 = 2000` over Modbus, then read the cloud `setting/get` for `ExportLimit`: **2000**. Same setting, not a mirror of it.

## u32, watts, clamped at 65535

- `46616` = high word, `46617` = low word. 46616 reads `0` here, which is why it is invisible if you only look at non-zero registers.
- Writing `[0, 3000]` starting at 46616 → both words update, cloud reports 3000.
- Writing 70000 (`[1, 4464]`) → clamped to 65535.
- Writing **0** works, so full export curtailment is possible. That is exactly what people on negative-price tariffs are asking for.

## Proof it limits actual feed-in

The battery was absorbing most of the PV, so I first set work mode to Feed-in First to push PV to the grid, then wrote a **1000 W** limit:

| | PV | load | feed-in | battery charge |
|---|---|---|---|---|
| +20s | 3.47 kW | 2.13 kW | **1.122 kW** | 0.166 kW |
| +40s | 4.09 kW | 2.08 kW | **1.117 kW** | 0.779 kW |
| +60s | 4.32 kW | 2.15 kW | **1.078 kW** | 0.982 kW |

PV rose 0.85 kW while feed-in stayed pinned at ~1.1 kW; the surplus went into the battery instead. Without the limit, feed-in tracks PV.

## Suggestion

Consider declaring it the same way `active_power` is — `[46617, 46616]`, low-word first, written as a u32 — rather than a single write to 41012, and keep it KH-only. I'm happy to test a branch on my unit.

Unrelated but possibly useful for the `CHARGE_PERIODS` table: on this KH10, registers **41001–41006 are readable but the inverter rejects writes** with a Modbus exception response (tested with both single- and multi-register writes), so excluding KH there looks correct.

PR #936, opened 2025-10-08
Correcting EPS power register for single-phase H1_G2s with newer firmwares

paultweedy77:
Post Manager v1.44 (and confirmed in later v1.49), EPS Combined Power is present on 39216/39217 (I32) for H1_G2 inverters, as per latest modbus spec as quoted in https://github.com/nathanmarlor/foxess_modbus/discussions/883#discussioncomment-14603388


MartB:
@paultweedy77 Would you mind picking this up again? Is there anything i can help you with? 
If you set-up the recommended dev-container in vscode you can just run
`pytest --snapshot-update` to fix the tests.

Issue #948, opened 2025-10-28
wrong H3 register used for invbatpower in remote_control_description.py


HerrKlaus:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### FoxESS - Modbus version

latest

### Inverter model

H3-12-E

### Inverter manager version

1.84

### Describe the bug

Intro:
I do not use your integration in HA as I'm using ioBroker, but I check from time to time some code changes here related to H3 registers.
Think I found a Code Bug during my readings.

Source of assumption:
entity_descriptions.py is stating around line 1428 "yield from _invbatpower" with H3 register (holding=[31036], models=Inv.H3_SET)

Issue1:
remote_control_description.py is using invbatpower=[31022] in line 103 for models=Inv.H3_SET & ~Inv.KUARA_H3 & ~Inv.AIO_H3_101 & ~Inv.AIO_H3_PRE101 -> at least for H3 this is not the right register, 31036 would be the right one

And another hint, not sure if this is a bug or needed in exactly this way:
entity_descriptions.py is stating in line 1431 & 1432:
ModbusAddressesSpec(input=[11008], models=Inv.H1_G1 | Inv.KH_PRE119),
ModbusAddressesSpec(holding=[31022], models=Inv.H1_G1 | Inv.H1_LAN | Inv.H1_G2_SET | Inv.KH_PRE133 | Inv.KH_133);
-> models=Inv.H1_G1 is listed as input and holding register -> is that intended?

nathanmarlor:
Fixed in #1168 — the H3 remote-control config was reading register 31022 (EPS Power R) instead of 31036 (battery power) for `invbatpower`. Now corrected. Merged to `main`; it'll be in the next release. Thanks @MartB for confirming. Closing.

nathanmarlor:
Quick update: this fix is now available in the **v1.16.0b1** beta — install it via HACS by enabling "show beta versions" and selecting v1.16.0b1, then restart. It'll roll into the next stable release after that.

Issue #669, opened 2024-09-02
EPS power moved in H3? 


canton7:
I did a quick EPS test, and I think EPS Power is in holding register 39217 (and possibly 39216 if it's a 32-bit value - which would make the H1_G2 the same as the H3_PRO, counting against one of the 3 phases I guess). It may always have been there, or this may be new as of Manager 1.44 - I've no way of testing that unfortunately.

_Originally posted by @paultweedy77 in https://github.com/nathanmarlor/foxess_modbus/discussions/666#discussioncomment-10520750_

canton7:
@paultweedy77 I'd find it very odd if just the EPS power has moved out to 39217 (and things like EPS current have stayed put). Sure the H3 uses that register, but as you say it uses it for the T phase.

I guess they might have moved it out if they wanted to make it a 32-bit register and needed more room... Apparently the G2 has a max capacity of 33kW, which just exceeds a 16-bit register. I would have expected them to move rpower as well though, in that case.

Are you certain that the EPS power has moved out to 39217, and nothing else EPS-related (or power-related) has moved?

paultweedy77:
The only thing I'm certain about is that register contains the correct EPS power figure when in EPS mode, and is 0 otherwise. My suspicion is that it's unintentional and shouldn't be relied on, and that the change in behaviour of register 31012 in Manager 1.44 is a bug.

nathanmarlor:
Review note for when this is tackled:

EPS Power (`eps_rpower`) for H1_G2/KH still reads holding **31012** (`entities/entity_descriptions.py`), which returns garbage on newer firmware (the 50000–60000 values that fail `Range(0,100)`). The correct register appears to be holding **39217/39216** (single-phase apparent power), reported as confirmed on H1-G2 (Manager 1.44+) and KH9.9 (Manager 1.54).

- PR #936 attempted exactly this (31012 → 39217 for H1_G2) but was **closed unmerged** — worth checking why before redoing (was it unconfirmed, or scoping?).
- This is **not** addressed by #1168, which fixed the H3 *battery* power register (31022 was EPS Power R; battery is 31036) — different register/model line.
- **#911** is the same root cause on the KH and should be fixed in the same change.

When you've confirmed 39217/39216 is right (and the firmware gating, if any), I'm happy to implement + regenerate snapshots.

wrcvr4:
I think I have the same issue in my H3_Smart, started with the beta 15, and now on the latest 16 beta
[...]

Issue #938, opened 2025-10-10
Force charge changes work mode to backup-up


mbedk:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### FoxESS - Modbus version

1.14.0b3

### Inverter model

H3-12.0-Smart

### Inverter manager version

1.16

### Describe the bug

I am trying to force charge using the provided work mode drop down, but it does not work and changes the work mode to back-up.

Not sure what is happening as it worked the first time but now it is no longer possible and trying to change the work mode back to Self Use does not work.

Only after some time or trying multiple times through the foxess app, the system seems to go back to self use.

Once the system is stable again, I seem to be able to force charge but only using the foxess app.

MartB:
https://github.com/nathanmarlor/foxess_modbus/blob/main/custom_components/foxess_modbus/remote_control_manager.py

Has a ton of back up mode references, most likely you are hitting one of these.

vava:
I have the same issue, looks like the modes are linked to the wrong values.

"Self Use" sets "Self Use" as expected.
"Feed-in First" can't be set, is changed back to "Self Use".
"Back-up" can't be set, is changed back to "Self Use".
"Peak Shaving" sets "Peak Shaving" as expected.
"Force Charge" sets "Backup" mode.
"Force Discharge" sets "Feed In" mode.

mbedk:
Hi @canton7 

I can confirm that the workaround described in https://github.com/nathanmarlor/foxess_modbus/discussions/1022 fixes the problem not being able to force charge on H3 Smart inverters. There are multiple issues and discussions open about this topic.

I am not a developer myself, however I would gladly help testing to get this issue solved.

Firespray-31:
I could be wrong, but the discussion and suggested workaround in #1022 suggests the problem lies with the `pwr_limit_bat_up` registers for the H3-Smart, and not the work mode registers?

Actually I think there is already a (w.i.p) PR #949 which fixes these registers (instead of just setting it to `None`) for the H3-Smart (and possibly some other registers for other inverters) that is not yet merged?

Specifically this commit: 

https://github.com/nathanmarlor/foxess_modbus/pull/949/changes/5450366067870f527af4352015b214edb35db4b8#diff-94e0ed579a391190aa25a3871ee9def49fbb1a77288dbdac399e2bf81d325279R140

itm1960:
[...]

PR #949, opened 2025-10-29
wip: A package of fixes for H3 incl. remote control and registers

MartB:
I need to rebase some changes i have done on my own fork so dont merge yet, but this will fix issues like: 

- [x] #948
- [ ] #362 #669 

**If people reading this would help and join #950, this would aid in making sure we mark registers correctly.**

### DIFF
--- a/custom_components/foxess_modbus/entities/modbus_battery_sensor.py
+++ b/custom_components/foxess_modbus/entities/modbus_battery_sensor.py
@@ -73,15 +73,15 @@ class ModbusBatterySensor(ModbusSensor):
 
     @property
     def native_value(self) -> Any:
-        if self._bms_connect_state_address is not None:
-            bms_connect_state = self._controller.read(self._bms_connect_state_address, signed=False)
-            # 0: Initial state, 1: OK, 2: NG
-            #
-            # On the H3, looks like this can be 247?
-            # https://github.com/nathanmarlor/foxess_modbus/issues/832#issuecomment-3183924877
-            # I'm going to assume that any values other than 0 / 2 mean "connected"...
-            if bms_connect_state == 0 or bms_connect_state == 2:
-                return None
+        value = (
+            self._controller.read(self._bms_connect_state_address, signed=False)
+            if self._bms_connect_state_address is not None
+            else None
+        )
+
+        # If the value is missing or BMS is disconnected, return None
+        if not value or (value & 0xFF) != 1:
+            return None
 
         return super().native_value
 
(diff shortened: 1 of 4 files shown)


Leon99:
Hey @MartB, how's your masters thesis? ;)

MartB:
@Leon99 had my defense last week and passed 💪🏻😂

However, I have had problems with H3 manager version 1.93 and therefore couldn't really develop anything since then.

Fox now downgraded me to 1.91 and I'm planning to take a look this Sunday. 1.93 is broken and the their devs are looking at it.

Leon99:
Congrats @MartB! 🎉 

I have H3 Smart, manager 1.26 (latest), let me know if I can help.

MartB:
Thanks! Sadly all of these changes are only for the legacy "H3" not for the smart versions. Is there any particular issue that you encounter that we don't track in issues already?

Leon99:
I came here from https://github.com/nathanmarlor/foxess_modbus/issues/938#issuecomment-4149675521

nathanmarlor:
Review notes for when this is picked back up (not merging — leaving for manual review):

**Already on `main`, can be dropped from this PR:**
- `invbatpower` 31022 → **31036** for H3 — merged in #1168.
- `battery_soc=[31038]` for H3 — already present.

**Needs hardware confirmation before merge:**
- The `pwr_limit_bat_up` → `pwr_limit_bat_down` rename plus the H3-Pro/Smart address change **46019/46018 → 46021/46020**. This overlaps with #979, which proposes the same 46021/46020 change but scoped to IVO only — please reconcile the two so H3_PRO/SMART and IVO end up with the correct (and consistent) registers/name.
- The H3 firmware split (`H3_PRE180` vs `H3_180`) giving H3 ≥ 1.80 `pwr_limit_bat_down=[44012]`. The `Inv.H3_180`/`H3_PRE180` flags exist, so it's wired up — just needs confirmation that 44012 is correct on H3 ≥ 1.80.

**Has a regression — must fix before merge:**
- In `modbus_battery_sensor.py`, the rewritten `native_value` does `value = read(addr) if addr is not None else None` then `if not value or (value & 0xFF) != 1: return None`. When `_bms_connect_state_address is None`, `value` is `None`, so it now returns `None` for **every** battery sensor that has no BMS-connect-state address — those sensors would all go unavailable. The old code only short-circuited when an address was present. Please keep the `if self._bms_connect_state_address is not None:` guard around the check.

**Also:** needs a rebase onto current `main` + `pytest --snapshot-update`.

Happy to help action any of these once you've decided on the register/firmware specifics.

PR #962, opened 2025-11-16
Change scale from 0.000001 to 0.001

### LINKED ISSUE #923: Wrong scale for Inverter Power?
### Is there an existing issue for this?

- [x] I have searched the existing issues

### FoxESS - Modbus version

latest

### Inverter model

H3-15 Pro

### Inverter manager version

1.66

### Describe the bug

It looks like Inverter Power is not scaled correctly. I'm new to Homeassistant, so a solution might be flipping a switch somewhere, but for now I'm stuck with weird graphs when comparing power consumption and generation.

The comment here suggests that mW is used https://github.com/nathanmarlor/foxess_modbus/blob/2b3107016bcc0cc2a563c0fb2dabcd573a2a1a9d/custom_components/foxess_modbus/entities/entity_descriptions.py#L868

But it looks to me like it's not.

Bildschirmfoto 2025-09-20 um 13 45 53

The values in the graphs should be up to 15kW.

phoet:
Updated scale factor for power measurement. Closes #923


nathanmarlor (dismissed):
Requesting changes before this can be merged:

1. **Evidence needed.** The original code had an explicit comment: "This one appears to be in mW, despite what the spec says" — indicating someone tested this and deliberately chose the 0.000001 scale. Please provide actual register readings from your inverter showing the current scale produces incorrect values and the proposed 0.001 produces correct ones.

2. **Missing snapshot updates.** The test snapshots for H3_PRO, H3_SMART, ENPAL_IX, 1KOMMA5, etc. still reference `1e-06`. Run the tests with `--snapshot-update` and include the updated files.

The consistency argument with R/S/T phase registers is compelling, but we need real hardware evidence before changing a value that was set intentionally.

jschmitz-de:
Can confim the Value is in "W" - same Scale as inv power R/S/T on 

> Master V: 1.74
> Manager V: 1.44
> Slave V: 1.06

PR #1107, opened 2026-04-23
Added support for F Series (G3)

liamjvs:
Added support for F Series G3 (F_G3) model handling in the integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Supported models:
 - FoxESS H3 (including AC3 and AOI-H3)
 - FoxESS H3 PRO
 - FoxESS KH
+- FoxESS F (G3)
 - Kuara H3
 - Sonnenkraft SK-HWR
 - STAR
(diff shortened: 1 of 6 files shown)


nathanmarlor:
Review notes (queued for a closer look):

- **`default_capacity_watts=0`.** `inverter_capacity` would then return 0, which feeds `native_max_value_callback=lambda x: x.inverter_capacity` for the remote-control / charge-power number entities — giving a 0 max. Since F_G3 is LAN-only and only exposes the 39xxx block, it likely doesn't enable those numbers at all; could you confirm that's the case? If it does expose any controllable numbers, we'd want a real capacity rather than 0.
- **Assertion relaxation.** Loosening the 'all profiles support AUX' assert to 'AUX *or* LAN' is sensible for a LAN-only model, but it's a global validation change — just confirming that's intended and doesn't mask a misconfiguration for other profiles.
- Otherwise this looks nicely scoped and the `F_G3-LAN` snapshot is included. Mainly need the capacity question answered before merge.

Thanks for adding F-series support!

PR #1116, opened 2026-04-30
Fix detection of H3-12.0-E (firmware 2.23 adds -E suffix)

vtainc:
Firmware 2.23 for H3 inverters started reporting model strings like "H3-12.0-E".
The existing regex ^H3-([\d\.]+) does not match this format, causing model
detection to fail and the integration to fall back to the serial number string.

This patch extends the regex to accept the optional "-E" suffix:

    ^H3-([\d\.]+)(?:-E)?

This change is backward‑compatible and does not affect older firmware versions.


vtainc:
But anyway the firmware is buggy, so this is only a workaround. There can be situations after inverter start that intergration will not start due to some incorrect values received, need to restart then inverter if this happens.

I have downgraded my inverter back to stable firmware.

PR #1132, opened 2026-05-18
Add Ebyte NA111 series as a supported network adapter

farajfarook:
Adds the Ebyte NA111 family of RS485-to-Ethernet serial servers to the
network adapter dropdown in the config flow.

The NA111, NA111-A, NA111-E and NA111-M share the same firmware and
protocol stack — they differ only in power input (DC 8–28 V, AC 85–265 V,
or PoE) and mounting form factor. All four support Modbus TCP and UDP
out of the box, so a single entry covers the whole series.

Configured with `max_read=100`, matching other capable RS485-to-Ethernet
adapters (Elfin EW11, Waveshare RS485-to-ETH, USR-TCP232-410s).

Tested on a FoxESS KH10 with the NA111-E over Modbus TCP.
Pre-commit hooks (ruff, ruff-format, prettier) pass locally.

Product page: https://www.cdebyte.com/products/NA111-E

### DIFF
--- a/custom_components/foxess_modbus/inverter_adapters.py
+++ b/custom_components/foxess_modbus/inverter_adapters.py
@@ -189,7 +189,7 @@ ADAPTERS = {
             config=_DefaultConfig(max_read=100),
         ),
         InverterAdapter.network(
-            "ebyte_na111_e",
+            "ebyte_na111_series",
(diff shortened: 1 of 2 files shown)


farajfarook:
@nathanmarlor what do you think mate? 👍🏽 or 👎🏽 ?

PR #1134, opened 2026-05-22
Fix EVO work mode and remote control; remove unverified charge periods

markybry:
## Summary

The initial EVO support (PR #1071) used incorrect registers for the EVO inverter's work mode and wired EVO into the H1-G2 charge-period registers, and the EVO had no remote-control configuration (so force charge/discharge was unavailable). This PR fixes work mode and adds remote control based on testing with actual EVO 10-H hardware, and **removes the unverified EVO charge-period support** (see rationale below).

### Work mode

- EVO reads work mode **1-based** (1=Self Use, 2=Feed-in First, 3=Back-up, 4=Peak Shaving) but **writes 0-based**
- Value **255** means the inverter is under external remote control (read-only state, not writable)
- A separate `write_map` is needed because the read and write encodings differ

### Remote control (force charge / discharge)

- The EVO had no `RemoteControlAddressSpec`, so no remote-control manager was created. This meant force charge/discharge and the Force Charge/Discharge Power numbers were never available.
- Added an EVO spec mirroring the H3-Pro / H3-Smart **46xxx** remote-control register family (which the EVO shares, along with work mode at 49203), with EVO-specific `battery_soc` (39423), `invbatpower` (39238/39237), `pv_voltages` (3 strings), and a 0-based work-mode fallback write map.
- Force charge/discharge is a **command, not an operating mode**, so for the EVO it is exposed as a dedicated **"Remote Control"** select (Disable / Force Charge / Force Discharge) rather than as extra options on the Work Mode select. The Work Mode select stays limited to the real work modes.

### Charge periods — removed (was unverified)

PR #1071 wired EVO into the H1-G2 charge-period registers (`Inv.H1_G2_SET | Inv.EVO`, the 41xxx range), which are not correct for EVO. While investigating the correct EVO registers (48xxx), it became clear the feature is only partly understood:

- EVO's "mode scheduler" sets a **per-slot work mode** (Self Use, Feed-in, Back-up, …), not a binary charge / no-charge toggle. Modelling it as charge-from-grid on/off is wrong.
- There appears to be a separate **master enable** that puts the inverter into scheduler mode, which isn't accounted for.
- Behaviour has only been observed on EVO 10-H, and only start/end times confirmed — the per-slot mode values and register layout across EVO variants are unverified.

Rather than ship (or "fix") a half-mapped feature, this PR **removes EVO from the charge-period spec entirely** so no misleading charge-period entities are created. It can return as a focused change once the registers are validated on hardware. This also means the EVO no longer touches the charge-period service at all.

### Infrastructure changes (needed to support EVO correctly)

- `ModbusWorkModeSelectDescription` gains an optional `write_map` for asymmetric read/write option encodings, and an `include_remote_control_modes` flag to keep force charge/discharge out of the Work Mode select
- `RemoteControlAddressSpec` gains a `dedicated_mode_select` flag to expose force charge/discharge as a standalone Remote Control select

There are **no charge-period service changes** in this PR — the previous draft's per-period write refactor has been dropped along with the charge-period support, so the shared charge-period service is untouched.

## Test plan

- [x] All existing snapshot and entity tests pass (`pytest tests/`)
- [x] EVO / EVO_10_H snapshots updated; no charge-period entities are created for EVO
- [x] Work mode 255 ("Remote Control") appears correctly when inverter is under external control
- [x] Validated on EVO 10-H hardware: standalone "Remote Control" select engages force charge/discharge; Force Charge/Discharge Power numbers present; Work Mode select limited to real work modes

> Note: the four remote-control-only registers (46001 remote_enable, 46002 timeout_set, 46004/46003 active_power, 46019/46018 pwr_limit_bat_up) are inferred from the H3-Pro/H3-Smart family that the EVO otherwise matches, since they are not exposed as standalone entities. Feedback from other EVO owners welcome.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


markybry:
The EVO exposes force charge/discharge via the remote control register family (46xxx), and register 49203 reads back 255 when the inverter is under external remote control, a read-only indicator, not a settable mode. This makes the EVO fundamentally different from other inverters where force
  charge/discharge are persistent work mode values you write directly.
  
Folding force charge/discharge into the Work Mode select conflates two different concepts: what the inverter does ordinarily (Self Use, Feed-in First, Back-up, Peak Shaving) versus what you are commanding it to do right now. Self Use already charges and discharges dynamically across the day. Force Discharge is an imperative override of that, not a peer operating strategy.

Keeping them in a separate Remote Control select makes the distinction clear in the UI, and it preserves the select.fox_inverter_remote_control entity key so existing dashboards and automations don't break.

markybry:
@nathanmarlor ay chance of getting some feedback on this?

james194zt:
@markybry  does this actually work in your code against your Evo? I am having issues getting scheduling working I just get illegal responses back from the modbus. If it is working may I ask what firmware yours is on? I emailed Fox trying to see if there is a newer modbus document that covers the Evo as the one around is over a year old but heard nothing back.

I have done a massive expansion of the modbus registers in my version for the Evo, basically everything you see in the app is now in the modbus, over 130 sensors now so I was going to PR in once complete, but I just cannot get this side of things working properly at all. Annoying as I have also written a Fox Plant plugin that allows you to fully replace the app with Home Assistant (I don't like China having access to my homes power), but I can't get my version of Stormsafe or its scheduling working over this modbus integration yet which is holding up my development here as well.

Thanks

markybry:
@nathanmarlor yes it does work as far as setting work modes , get sensor readings back and controlling the device. 

Currently I am using the integration to Force Discharge during peak times. Tuning the feed-in priority for sunny days to use my battery most effectively and setting the night charge to optimise export during the peak period the next day based on how much predicted PV production will happen during the next day.

However the charge period may well have broken pulling in your latest changes. Something I did not retest. Let me work on that and update the PR.

Here are my firmware details.
Version_Master：
1.15
Version_Slave：
1.00
Version_Manager：
1.10
Version_BCU：
1.004
Version_AFCI：
0.35

markybry:
# FoxESS EVO — getting charge periods / scheduling working over Modbus

Verified working against an **EVO 10-H**, firmware **Master 1.15 / Manager 1.10 / Slave 1.00**.
Both periods read and write reliably — no illegal-function / illegal-data-address responses in the logs.

## TL;DR — why you're getting illegal responses

The EVO does **not** use the H1/H3 `41xxx` charge-period registers. Writing those (or any address the
firmware doesn't expose) returns **illegal data address (exception 02)**. The EVO has its own block at
`480xx` and an **extra per-period "mode" register** that isn't in the public Modbus doc.

> **Note:** this integration currently maps **two** charge periods. Whether the EVO firmware exposes
> *more* slots (e.g. `48030+`) is **unverified** — I've only confirmed periods 1 and 2 work. See
> [Charge periods vs the app Scheduler](#charge-periods-vs-the-app-scheduler-stormsafe) below if you're
> after full app-replacement scheduling, which is almost certainly a different register block.

## Register map

| | enable charge-from-grid | start | end | **mode** |
|---|---|---|---|---|
| **Period 1** | `48010` | `48011` | `48012` | **`48013`** |
| **Period 2** | `48020` | `48021` | `48022` | **`48023`** |

- All holding registers.
- `start` / `end` use the usual `(hour << 8) | minute` time encoding.
- `enable charge-from-grid`: `1` / `0`.
- **`mode`** (the bit that unlocked it): **`6` = charge**, **`1` = no-charge**. Set it per period
  alongside the times.

## Three things to get right

1. **Use the 480xx block, not 41xxx.** The EVO was initially (incorrectly) mapped onto the H1_G2
   `41xxx` registers — that's an illegal address on EVO hardware.

2. **Write the mode register.** Writing start/end/grid-enable without `48013` / `48023` won't behave
   correctly. Mode follows the grid-charge flag: `6` when charging, `1` otherwise.

3. **Write each period as its own contiguous block.** The two periods live in **separate** ranges
   (`48010–48013` and `48020–48023`), so you can't assume one contiguous span across all periods and do
   a single write. Do one FC16 block write per period, sized `max(addr) − min(addr) + 1`.

## How many periods?

This integration maps **two** charge periods (the `480xx` block above), and the write machinery is
driven entirely by that definition list — it loops over however many periods are defined and writes
each as its own block. Two things are hardcoded to 2, though:

- `CHARGE_PERIODS` defines exactly two periods.
- The service schemas cap it: `update_charge_period` accepts `charge_period` in range **1–2**, and
  `update_all_charge_periods` requires **exactly 2** entries.

To support more you'd add entries to `CHARGE_PERIODS` (with their register addresses) and loosen both
schema ranges — the rest of the plumbing is count-agnostic.

**Unverified:** whether the EVO firmware actually exposes more than 2 charge-period slots (e.g.
`48030–48033`). I've only confirmed periods 1 and 2. If your newer "list of periods" code writes a 3rd+
slot to addresses the firmware doesn't implement, that would return **illegal data address** — worth
checking whether that's what you're hitting.

## Charge periods vs the app Scheduler (Stormsafe)

Don't conflate these — they're separate features:

- **Charge periods** (this doc, `480xx`): the legacy 2-slot force-charge / charge-from-grid windows.
- **Scheduler / Stormsafe** (the app's "Scheduler"): up to ~8 time segments, each with its own work
  mode, min-SoC, etc. This lives in a **different, larger register block** — *not* the `480xx`
  registers here.

If the goal is full app replacement (driving the inverter entirely from Home Assistant), the 8-segment
Scheduler block is what you need, and it's unrelated to the charge-period registers documented here.

## Reference diff

### `entities/charge_period_descriptions.py`
Split EVO out of `Inv.H1_G2_SET` and point it at the 480xx block:

```python
ChargePeriodAddressSpec(
    holding=ModbusChargePeriodAddressConfig(
        period_start_address=48011,
        period_end_address=48012,
        enable_charge_from_grid_address=48010,
        mode_address=48013,
    ),
    models=Inv.EVO,
),
# ...and for period 2: 48021 / 48022 / 48020 / mode 48023
```

### `entities/modbus_charge_period_config.py`
Add the mode register to the address config:

```python
@dataclass(frozen=True)
class ModbusChargePeriodAddressConfig:
    period_start_address: int
    period_end_address: int
    enable_charge_from_grid_address: int
    mode_address: int | None = None      # EVO only
    mode_charge_value: int = 6
    mode_no_charge_value: int = 1
```

### `services/update_charge_period_service.py`
Write each period separately so non-contiguous ranges (EVO) work, and include the mode register:

```python
# Write each charge period separately so non-contiguous address ranges (e.g. EVO) are handled correctly
for charge_period, config in zip(charge_periods, controller.charge_periods, strict=True):
    writes: list[tuple[int, int]] = [
        (config.addresses.period_start_address,
         serialize_time_to_value(charge_period.start) if charge_period.enable_force_charge else 0),
        (config.addresses.period_end_address,
         serialize_time_to_value(charge_period.end) if charge_period.enable_force_charge else 0),
        (config.addresses.enable_charge_from_grid_address,
         1 if charge_period.enable_charge_from_grid else 0),
    ]
    if config.addresses.mode_address is not None:
        mode_value = (config.addresses.mode_charge_value
                      if charge_period.enable_charge_from_grid
                      else config.addresses.mode_no_charge_value)
        writes.append((config.addresses.mode_address, mode_value))

    write_start_address = min(w[0] for w in writes)
    write_end_address = max(w[0] for w in writes)
    write_values: list[int] = [None] * (write_end_address - write_start_address + 1)
    for address, value in writes:
        write_values[address - write_start_address] = value

    await controller.write_registers(write_start_address, write_values)
```

## Live read-back (proof it works)

| | Start | End | Force charge | Charge from grid |
|---|---|---|---|---|
| Period 1 | 16:00 | 19:00 | on | on |
| Period 2 | 16:30 | 17:15 | off | off |

nathanmarlor:
Review notes (queued for a closer look):

- **Confirm the charge-period regression is resolved.** You mentioned earlier that pulling in the latest changes may have broken the charge period — could you confirm whether that's fixed and that force charge/discharge + the charge-period writes are working on your EVO now?
- **Shared-infra blast radius.** This touches `ModbusChargePeriodAddressConfig.mode_address`, `ModbusWorkModeSelectDescription.write_map`/`include_remote_control_modes`, `RemoteControlAddressSpec.dedicated_mode_select` and the per-period charge-period writes — all shared with H1/H3/KH. The main thing to verify before merge is that these don't change behaviour for non-EVO models (the snapshots help, but the work-mode write path especially).
- **EVO remote-control registers.** Some of the RC registers (46001/46002/46004/46019 etc.) appear to be inferred from the H3-Pro/Smart family rather than confirmed on EVO hardware — worth a sanity check against a real EVO if possible.
- Will need a **rebase onto current `main` + `pytest --snapshot-update`** (a lot has merged recently).

Functionally the structure looks reasonable; these are the points to clear before merging.

markybry:
Thanks @nathanmarlor — going through your points:

**1. Charge-period regression.** Fixed. The breakage was the old single combined-block write; `_set_charge_periods` now writes each period as its own contiguous block, which is what the EVO's non-adjacent `480xx` ranges need. The read-back table above (P1 16:00–19:00 charge-on, P2 16:30–17:15 off) is from a live write/read cycle on the EVO-10H — no illegal-data-address responses. Force charge/discharge is solid too; I run force-discharge daily during peak.

**2. Shared-infra blast radius.** EVO-specific behaviour is all behind defaults so non-EVO is untouched: `mode_address` defaults `None` (no mode write for H1/H3/KH), and `include_remote_control_modes=True` / `dedicated_mode_select=False` / `write_map=None` keep the work-mode and RC selects identical for every non-EVO model — the only snapshot diffs are the two EVO files.

The one genuinely shared change to flag is the charge-period **write path**: it now issues one FC16 per period instead of a single combined write across both. For H1/H3 that turns the old 6-register write (`41001–41006`) into two 3-register writes (`41001–41003`, `41004–41006`) — same values, same registers, but two transactions. That's the bit worth a sanity check on an H1/H3 if you have one to hand.

**3. EVO remote-control registers.** Correct — the `46xxx` family (`46001` enable, `46002` timeout, `46003/46004` active power, `46610` max-soc, `46018/46019` bat charge-power limit) is carried over from the H3-Pro/Smart spec, not an EVO doc. Hardware-confirmed: remote-enable + active-power force charge/discharge works on my EVO-10H. The SoC-limit registers (`46610`, `46019`) are inferred and not independently bench-verified — happy to mark them provisional in comments, or drop them until confirmed if you'd rather not ship unverified writes.

**4. Rebase + snapshots.** Will do — rebase onto current `main` and `pytest --snapshot-update`.

Firmware for reference: Master 1.15 / Manager 1.10 / Slave 1.00, EVO-10H.

markybry:
Heads-up on a scope change in the latest force-push (rebased onto current `main`, `441a22f`):

**I've removed the EVO charge-period support from this PR.** The original EVO PR (#1071) wired EVO into the H1-G2 charge-period registers (41xxx); my earlier draft "corrected" those to 48xxx. On closer inspection the feature is only half-mapped:

- EVO's mode scheduler sets a **per-slot work mode**, not a binary charge/no-charge — so the existing `enable_charge_from_grid` model is semantically wrong for EVO.
- There's a separate **master "scheduler enable"** that isn't handled, so writing slots may do nothing on its own.
- Only the 10-H has been observed, and only start/end times confirmed; per-slot mode values and the layout across EVO variants are unverified.

Rather than ship something I can't stand behind, I've dropped EVO from the charge-period spec entirely (just removes `| Inv.EVO` from the two H1-G2 specs). It can come back as a focused PR once I've mapped the registers on hardware.

Practical upshot for your review:

- **The per-period charge-period write refactor is gone.** That was the one non-EVO change you flagged as blast radius — the shared `update_charge_period_service.py` and `ModbusChargePeriodAddressConfig` are now byte-for-byte upstream.
- What remains is EVO-only: work-mode asymmetric read/write encoding (+ the 255 remote-control state), the standalone Remote Control select for force charge/discharge, the `battery_soc` (39423) fix, and the `EVO_10_H` model variant.
- Snapshots regenerated; EVO no longer emits any charge-period entities.

Net diff is now 10 files, all EVO-scoped except the 2-line `| Inv.EVO` removal in `charge_period_descriptions.py`.

markybry:
@nathanmarlor did have any thoughts on how to get this PR moving again?

PR #1147, opened 2026-06-06
Update entity_descriptions.py _total_yield_today

jschmitz-de:
_total_yield_today may exceed 100 kWh.
The value is capped when "validate=[Range(-100, 100)]," is set


jschmitz-de:
i´ve tested this on my H3-Pro 20 and works fine. No more "validation errors" can be found in the log

PR #1245, opened 2026-08-30
Add charge period support for H3 Smart / H3 Pro (time-group system)

fabian1512:
## Summary

The charge period card works on H1/G2/EVO via the `41001..41006` charge period registers. The **H3 Smart** (and H3 Pro, which shares the same register map) has those registers marked invalid and instead defines charge periods with the **time-group system** (Table 3-11, base `48010 + 10*(N-1)`). This PR adds time-group charge period support for `Inv.H3_PRO_SET | Inv.H3_SMART`.

## Changes

- **`entities/modbus_charge_period_config.py`**
  - `ModbusChargePeriodAddressConfig`: added optional time-group fields (`enable_address`, `work_mode_address`, `soc_address`, `fdsoc_address`, `fdpwr_address`, `enable_flag_address`). `enable_address` doubles as the base of the 10-register block.
  - `ChargePeriodAddressSpec`: getters for the new fields; `_get_address` tolerates `None` (H1 configs unaffected).
  - `ModbusChargePeriodFactory`: new option to derive the force-charge binary sensor from the group enable register instead of the start/end times.
- **`entities/charge_period_descriptions.py`**: factories for H3 groups 1 and 2 (the same groups the FoxESS app uses) with Work Mode 6 (Force Charge) and FDPWR (`base+6`) as the "charge from grid" register (also the marker entity the card uses to discover supported inverters).
- **`entities/modbus_entity_mixin.py`**: `get_entity_id` fallback now includes the platform domain, so the charge period card can resolve entities before they are registered.
- **`services/update_charge_period_service.py`**: for time-group inverters, write the full group block (enable, start/end, Work Mode 6, SoC, FDPWR, enable flag) and tie the Time Period Mode flag (`48000`) to whether any period is enabled. The H1 write path is preserved (writes are split into contiguous runs).

## Testing

- Verified live on an H3 Smart (`Foxy`): Period off → normal/discharge; Period on + grid off → PV-only force charge; Period on + grid on → grid charging.
- `python -m pytest`: 35 passed (snapshots regenerated for the models mapping to `Inv.H3_SMART` / `Inv.H3_PRO_SET`).
- `ruff`, `ruff format --check`, `mypy`: clean.

## Notes / limitations

- The device/app must be in "Modus Terminierer" (the time-period mode) for the groups to run; this is the app/Cloud level and not set by the card.
- Force-charge power defaults to the inverter capacity (grid on) and SoC bounds default to 100/10, as the card doesn't expose these.

### DIFF
--- a/custom_components/foxess_modbus/entities/charge_period_descriptions.py
+++ b/custom_components/foxess_modbus/entities/charge_period_descriptions.py
@@ -9,6 +9,63 @@ from .modbus_charge_period_config import ModbusChargePeriodFactory
 
 _LOGGER: logging.Logger = logging.getLogger(__package__)
 
+
+# The H3 Smart defines its charge periods via the time-group system (Table 3-11).
+# Each group is a contiguous block of 10 registers, starting at 48010 + 10*(N-1).
+# We use groups 1 and 2, which are the same groups the FoxESS app uses for its charge periods, so that
+# the card and the app stay in sync. Unlike the H1, force charge is controlled by a dedicated enable
+# register (the group's enable), rather than being inferred from the start/end times, so these factories
+# use enable_force_charge_from_enable_address=True.
+def _h3_group_base(group: int) -> int:
+    """Returns the base address of the given time group (1-based)."""
+    return 48010 + 10 * (group - 1)
+
+
+def _h3_charge_period_factory(
+    group: int,
+    period_start_key: str,
+    period_start_name: str,
+    period_end_key: str,
+    period_end_name: str,
+    enable_force_charge_key: str,
+    enable_force_charge_name: str,
+    enable_charge_from_grid_key: str,
+    enable_charge_from_grid_name: str,
+) -> ModbusChargePeriodFactory:
+    base = _h3_group_base(group)
+    return ModbusChargePeriodFactory(
+        addresses=[
+            ChargePeriodAddressSpec(
+                holding=ModbusChargePeriodAddressConfig(
+                    period_start_address=base + 1,
+                    period_end_address=base + 2,
+                    # 'Enable charge from grid' is determined by whether the force-charge power (FDPWR,
+                    # offset +6) is non-zero. We point this at the FDPWR register so that the resulting
+                    # binary sensor (device class power) also serves as the marker entity that the
+                    # charge-period card uses to discover supported inverters.
+                    enable_charge_from_grid_address=base + 6,
+                    enable_address=base,
+                    work_mode_address=base + 3,
+                    soc_address=base + 4,
+                    fdsoc_address=base + 5,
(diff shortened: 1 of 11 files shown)


fabian1512:
**Summary**

The charge period card currently supports H1/G2/EVO via the `41001..41006` registers. The **H3 Smart** (and H3 Pro, same register map) marks those invalid and instead defines charge periods with the **time-group system** (Table 3-11). This PR adds support for `Inv.H3_PRO_SET | Inv.H3_SMART`.

**How it works**

- Uses **Groups 1 + 2** (base `48010 + 10*(N-1)`) — the same groups the FoxESS app uses for its charge periods, so the card and the app stay in sync.
- Force charge = **Work Mode 6**; "charge from grid" is **FDPWR > 0** (we point the binary sensor at FDPWR — it also serves as the `device_class: power` marker the card uses to discover supported inverters).
- `48000` (Time Mode Flag) = master switch: `1` when any period is enabled, `0` when all are off (so the inverter returns to normal when everything is disabled).
- The group's SoC bounds (offset +4) are mirrored from the global Max SoC / Min SoC OnGrid (`46610`/`46611`, the GUI-editable values) rather than hardcoded, matching the app.

**Key finding for reviewers**

"Timer / Modus Terminierer" in the app is register **`48000`** (Time Mode Flag), **not** a 49203 Work Mode value — `49203` stays Self Use. The device/app must be in its "Timer" mode for the groups to run; the card can only control the local inverter state, not the app/Cloud mode.

**Files**

- `modbus_charge_period_config.py`: optional time-group fields; factory option to derive force-charge from the group enable register.
- `charge_period_descriptions.py`: H3 factories (Group 1+2).
- `modbus_entity_mixin.py`: `get_entity_id` fallback now includes the platform domain (needed for the card to resolve entities).
- `update_charge_period_service.py`: writes the full 10-register group block + `48000`; H1 write path preserved (contiguous runs).

**Testing**

`pytest` 35 passed (snapshots regenerated for the H3-family models); `ruff`/`format`/`mypy` clean; live-verified on an H3 Smart (off→normal/discharge, on+grid off→PV-only, on+grid on→grid charge). Battery SoC near MaxSoC correctly prevents charging.

**Caveats**

- The inverter resets repeated Modbus writes; read via `read_registers`/HA, write once + wait.
- H1's "hold" semantics (grid off = neither charge nor discharge) don't map cleanly to H3 (no hold work mode); grid-off maps to PV-only (Work Mode 6 + FDPWR=0).
- Force-charge power defaults to the inverter capacity; SoC bounds now come from the configured registers.

**Open questions**

- Is using Groups 1 + 2 acceptable (they overlap the app's own periods), or should a hidden group be preferred?
- Should `46620` (Max SoC From Grid) be exposed as a separate entity?

Issue #1124, opened 2026-05-07
Integration broken with update 2.23/1.93 (H3)


partxxx:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### FoxESS - Modbus version

v1.15.0b1

### Inverter model

H3-5.0-E

### Inverter manager version

1.93

### Describe the bug

Version_Master：2.23
Version_Slave：1.03
Version_ARM：1.93
Installation method Home Assistant OS
Core 2026.5.0
Supervisor 2026.04.2
Operating System 17.3
Frontend 20260429.3

I updated the inverter yesterday, and most of the values are now unknown or false.
I assume the Modbus table or the protocol changed with the update.

andre8777:
Same with my inverters , a few  Registers have changed like PV Power, Battery Power.

MartB:
I think there's just a bug with modbus in general on this version. I can't properly read registers that work on a second device with 1.93.

But yes they started to align registers with the H3 pro spec apparently. Not sure if by accident.

partxxx:
Yeah, this is not a single problem, in general the modbus integration changed, and the registers too.

domerich:
yeah and down grade does not work either

domerich:
I opened a ticket with FoxESS and they downgraded my inverter back to 2.22 / 1.92 and now everything works agin. Won't touch it as this repo seems dead.

partxxx:
> I opened a ticket with FoxESS and they downgraded my inverter back to 2.22 / 1.92 and now everything works agin. Won't touch it as this repo seems dead.

Thx, and what is yout HA / integration versions?

MartB:
There's nothing to fix in this repo for H3 manager 1.93. Modbus is buggy on 1.93 and fox confirmed it to me in a support ticket.

skipgub:
A bunch of our sensors had stopped working suddenly two days ago. Eventually, having exhausted all other avenues, I finally got around to updating the firmware on our AC1-5.0-E-G2-WL  inverter - and everything in this integration immediately started working again. FWIW we're now on Master 1.53, Manager 1.39 (and Slave unchanged 1.04).

domerich:
> > I opened a ticket with FoxESS and they downgraded my inverter back to 2.22 / 1.92 and now everything works agin. Won't touch it as this repo seems dead.
> 
> Thx, and what is yout HA / integration versions?

I’m using the 1.5 beta

[...]

PR #1253, opened 2026-09-10
Support H3 Manager >= 1.93 (Master 2.23): repair frames missing the unit id, add H3_193 profile

jrx-code:
## Summary

FoxESS H3 firmware **Manager 1.93+** (pushed with Master 2.23, see #1124) breaks this integration behind Modbus-TCP<->RTU gateways. This PR makes it work again and adds a firmware profile for it.

Measured on an **H3-10.0-E, Master 2.23 / Slave 1.03 / Manager 1.95**, AUX RS485, USR/Waveshare RS485-to-TCP gateway in "Modbus TCP<=>Modbus RTU" mode, slave 247, integration configured as "network / other adapter / tcp".

### 1. What the firmware does wrong

Roughly 90 % of RTU responses leave the inverter **without the slave-address byte**. The gateway forwards what it received, so the Modbus-TCP frame is one byte short:

```
good:  tid pid 0009 f7 03 06 00df 0067 0195     (unit f7, fc 03, 6 bytes: 223/103/405)
short: tid pid 0008    03 06 00df 0067 0195     (MBAP length 8, no unit id)
```

pymodbus takes `03` as the unit id and `06` as the function code, `_validate_slave_id` rejects it and the integration logs `No Response received from the remote slave/Unable to decode response`. The register contents inside the short frames are correct. Exceptions arrive the same way (`0002 83 02`).

### 2. Frame repair (`client/frame_repair.py`, `CustomModbusTcpClient`)

`MissingUnitIdRepairer` re-inserts the unit id **only** when byte 6 differs from the unit we addressed **and** equals the function code we sent (or `0x80 | fc`). A healthy device never matches: byte 6 equals the unit and is passed through, also for a slave id that happens to equal the function code (test included). pymodbus reads the 8-byte header and the body in two `recv()` calls, so the client keeps its own receive buffer; `execute()` records the pending unit/fc, `send()` re-arms for retries. Repair counters are logged at debug level.

Unit tests use frames captured on the inverter (`tests/test_frame_repair.py`).

### 3. Register survey with the repaired client

Every register the H3 profile polls reads correctly, individually and in blocks of up to 100 registers: 30000-30019, 31000-31135, 32000-32023, 41000, 41007-41019, 44000-44015. Only **41001-41006** (charge periods, not used by the H3 profile) return exception 02. The complete H3-Pro map (36001-49203) also answers on this firmware.

FC 0x06 writes round-trip correctly (41011 min SoC on grid 10 -> 11 -> 10, read back on the next poll; 41000 work mode).

### 4. `Inv.H3_193` profile

`battery_charge_today` (32005) and `battery_discharge_today` (32008) no longer hold today's energy. They hold `(total mod 65536) + 1` in 0.1 kWh units:

| registers | value | meaning |
|---|---|---|
| 32003, 32004 | 1, 32669 | battery charge total = 98205 -> 9820.5 kWh |
| 32005 | 32670 | "today" = 98205 - 65536 + 1 |
| 32006, 32007 | 1, 26432 | battery discharge total = 91968 -> 9196.8 kWh |
| 32008 | 26433 | "today" = 91968 - 65536 + 1 |

The H3-Pro style 32-bit registers 39607/39608 and 39611/39612 carry the same value x10, so no register holds the daily figure any more (the cloud reports 19.0 / 5.4 kWh for the same day). The existing `Range(0, 100)` validation already left both sensors permanently `unknown`, so the new profile simply does not create them.

- `Inv.H3_193` is part of `H3_SET` and is the `latest` for the H3 profile; `Version(1, 92)` keeps `H3_180` for older firmware (same pattern as the 1.80 split). Users on Manager < 1.93 who have "latest" selected can pick their version in the options flow, exactly as before.
- The six `H3_180`-only registers (31090, 31102, 31103, 31123, 31134, 31135) read fine on 1.95 and apply to `H3_193` too.
- `AC3` is left unchanged: I have no AC3 to measure.
- Snapshots: `H3-AUX-v1.92` added (identical to the previous `latest`), `H3-AUX-latest` minus the two entities.

### Verification

- 46 tests pass (`pytest`), `ruff` 0.9.4 and `mypy` clean.
- Running on HA 2026.9.1 for several hours: `connection_status` stays `Connected`, 83 entities with plausible values, totals equal the FoxESS cloud figures (grid consumption 16838.8 kWh, feed-in 2993.1 kWh, load 30994 kWh).

### Not covered

- Remote control writes to 44002-44013 with FC 0x10: a comment in #1124 reports that 1.93+ accepts only single-register writes there. Not tested here (it changes inverter behaviour).
- With two Modbus-TCP clients on the same gateway the gateway hands RS485 answers to either socket; this is unrelated to the fix but easy to hit while testing.


hamster65:
Thanks, can't wait to update....  I wonder if Fox Ess is aware of this bug?

